PR #1, opened 2018-07-23
Initial Creation of document-generator repo and maven module document-generator-interface

Mathsh:
Initial commit to set up the document-generator repo with MakeFile, Start.sh and Pom.xml
Initial Creation of document-generator-interface maven module with specific interfaces and docDataTpe Enum.

Alterations may be made as document-generator-core and document-generator-accounts maven modules are created.

Resolves SFA-425


document-generator-interface/pom.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.companieshouse</groupId>
```
jsimons1989: This should be: uk.gov.companieshouse

document-generator-interface/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+    <version>unversioned</version>
+  </parent>
+  <groupId>uk.gov.companieshouse</groupId>
+  <artifactId>document-generator-interface</artifactId>
+  <version>unversioned</version>
+  <name>document-generator-interface</name>
+  <url>http://maven.apache.org</url>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
```
jsimons1989: Can you add spacing to make this easier to read please? This will get difficult to read as it gets longer.

document-generator-interface/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>3.8.1</version>
```
jsimons1989: This is an old version. Some of our repos are using JUnit 5. Version 3 doesn't support annotations (e.g. @Test)

Makefile
```diff
@@ -0,0 +1,50 @@
...
+.PHONY: build
+build: submodules
+	mvn package
+	# Moving JAR to repo root so the path relative to the start script is the same in dev as it is in the build artefact
+	mv ./target/$(artifact_name).jar ./$(artifact_name).jar
+
+.PHONY: package
+package:
+	$(eval tmpdir:=$(shell mktemp -d build-XXXXXXXXXX))
+	cp ./$(artifact_name).jar $(tmpdir)
```
jsimons1989: This is going to fail to build. The JARs will be in the submodules.

document-generator-interface/pom.xml
cburns93: I agree, please make it 4 spaces instead of 2!

document-generator-interface/pom.xml
cburns93: I agree here also, please use either 4 or 5. I would recommend 5 as it follows the version our `company-accounts` repos are using.

document-generator-interface/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+      <scope>test</scope>
+    </dependency>
```
cburns93: There seems to be no artifactories specified here?

document-generator-interface/src/main/java/uk/gov/companieshouse/document/generator/DocumentDataFactory.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.companieshouse.document.generator;
+
+import uk.gov.companieshouse.document.generator.docdatatype.DocumentDataType;
+
+public interface DocumentDataFactory {
+    public GetDocumentInfo getDocumentInfo(DocumentDataType documentDataType);
```
cburns93: `public` is redundant in an interface

pom.xml
```diff
@@ -0,0 +1,30 @@
...
+
+    <properties>
+      <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <dependencies>
+      <dependency>
+        <groupId>junit</groupId>
+        <artifactId>junit</artifactId>
+        <version>3.8.1</version>
```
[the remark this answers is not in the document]
cburns93: I would use 5 to match our new company accounts repos

start.sh
```diff
@@ -0,0 +1,30 @@
...
+# Start script for document-generator
+
+APP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+if [[ -z "${MESOS_SLAVE_PID}" ]]; then
+    source ~/.chs_env/private_env
+    source ~/.chs_env/global_env
+    source ~/.chs_env/document-generator/env
+
+    PORT="${DOCUMENT_GENERATION_SERVICE_PORT:=4017}"
```
cburns93: We shouldn't be assigning the `4017 port. Because when we check in the code if that PORT is null it should throw an error not start on a random port.

document-generator-interface/src/main/java/uk/gov/companieshouse/document/generator/DocumentInfoFactory.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.companieshouse.document.generator;
+
+import uk.gov.companieshouse.document.generator.core.document.doctype.DocumentInfoType;
+
+public interface DocumentInfoFactory {
+    public GetDocumentInfo getDocumentInfo(DocumentInfoType documentDataType);
```
cburns93: Don't need redundant `public` access modifier

pom.xml
```diff
@@ -0,0 +1,29 @@
...
+
+    <properties>
+      <!-- Encoding -->
+      <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+      
+      <!-- JDK Version -->
+        <jdk.version>1.8</jdk.version>
+    </properties>
+
+    <dependencies>
```
cburns93: Can remove `dependencies` tag if there are none.

start.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+#
+# Start script for document-generator
+
+APP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+if [[ -z "${MESOS_SLAVE_PID}" ]]; then
+    source ~/.chs_env/private_env
+    source ~/.chs_env/global_env
+    source ~/.chs_env/document-generator/env
```
cburns93: There is no port configured here? When I said to take it out I mean to just do: `PORT="${DOCUMENT_GENERATION_SERVICE_PORT}"` rather than `PORT="${DOCUMENT_GENERATION_SERVICE_PORT:=4017}"` as the second says assign `DOCUMENT_GENERATION_SERVICE_PORT` to `PORT` however if it is null or doesn't exist, then assign `4017` as the default value.

document-generator-interface/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+    <parent>
+        <groupId>uk.gov.companieshouse</groupId>
+        <artifactId>document-generator</artifactId>
+        <version>unversioned</version>
+    </parent>
+  
+    <groupId>uk.gov.companieshouse</groupId>
+    <artifactId>document-generator-interface</artifactId>
+    <version>unversioned</version>
+    <name>document-generator-interface</name>
```
cburns93: Could we add a `description` tag? Just to say what it does? As a short description so when it gets built, the manifest for the jar has a nice description :)

Mathsh:
Pull request split to 

https://github.com/companieshouse/document-generator/pull/2
https://github.com/companieshouse/document-generator/pull/3

PR #6, opened 2018-08-14
Avro Model creation and serialiser/deserializer for document generation

Mathsh:
Created the following elements for Avro, to be used within the document-generator-core module in document-generator

Created Avro Serializer, Deserializer.
Created DocumentGenerationStateAvroSerializer to handle completed, failed and started serialised messages

Added Avro schema class generator plugin to pom.xml
Added sonar to pom.xml
Added folder called resources/avro to include the Avro schemas. 
Auto generated the Avro models from the schemas

Resolves SFA  421

PR #7, opened 2018-08-14
renderDocument response and request models

Mathsh:
created models for renderDocumentResponse and renderDocumentRequest

resolves SFA 421


document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/render/RenderDocumentReponse.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.companieshouse.document.generator.core.document.render;
+
+public class RenderDocumentReponse {
+
+    private DocumentGenerationCompleted generatedDocument;
```
ch-mcabrera: This class does not exist in the project nor its dependency in the pom.xml

Mathsh: It is present in https://github.com/companieshouse/document-generator/pull/6 which needs to be PRd first. I should have held the PR off on this one.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/render/RenderDocumentRequest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private String contentType;
+
+    private String documentType;
+
+    private String location;
+
+    private String data;
+
+    public RenderDocumentRequest(){}
```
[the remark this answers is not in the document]
Mathsh: I added an constructor with all params, which in turn means that Java does not add the default constructor. So I add one myself. Unless this has changed?

PR #11, opened 2018-08-16
Adds skeleton retrieval of Transaction functionality to AccountsDocumentInfoServiceImpl

ChrisJBurns:
- adds retrieval of transaction functionality
- adds `TransactionManager` that calls the `transactions-api` through `ERIC` to get the correct transaction data
- adds service layer that calls the `TransactionManager`

**Note:** It's important to note that the `TransactionManager` will be removed once the `private-sdk` is implemented. When that has been done, then the `TransactionService` simple uses the `private-sdk` rather than the internal project `TransactionManager` as the purpose of the `TransactionManager` is simply to mimic what the `private-sdk` does. Additionally, it is important to note that there are no tests for the `TransactionService` as it cannot test the `TransactionManager` due to static methods. When the `private-sdk` has been implemented, the `TransactionService` can then be tested. Until then, there will be empty tests.

Implements: SFA-567

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
@@ -27,6 +27,7 @@ public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
 
         Transaction transaction = transactionService.getTransaction(resource);
         if (transaction == null) {
+            LOG.error("transaction not found");
             return null;
         }
 
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
@@ -1,7 +1,7 @@
 package uk.gov.companieshouse.document.generator.accounts;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
+import org.springframework.stereotype.Service;
 import uk.gov.companieshouse.api.model.transaction.Transaction;
 import uk.gov.companieshouse.document.generator.accounts.service.TransactionService;
 import uk.gov.companieshouse.document.generator.interfaces.DocumentInfoService;
@@ -9,7 +9,7 @@ import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfo;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 
-@Component
+@Service
 public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
 
     @Autowired
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
@@ -1,12 +1,36 @@
 package uk.gov.companieshouse.document.generator.accounts;
 
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.model.transaction.Transaction;
+import uk.gov.companieshouse.document.generator.accounts.service.TransactionService;
 import uk.gov.companieshouse.document.generator.interfaces.DocumentInfoService;
 import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfo;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
 
+@Component
 public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
+
+    @Autowired
+    private TransactionService transactionService;
+
+    public static final String MODULE_NAME_SPACE = "document-generator-accounts";
+
+    private static final Logger LOG = LoggerFactory.getLogger(MODULE_NAME_SPACE);
 
     @Override
     public DocumentInfo getDocumentInfo() {
+        LOG.info("Started getting document");
+
+        String resource = "";
+
+        Transaction transaction = transactionService.getTransaction(resource);
+        if (transaction == null) {
+            return null;
+        }
+
         return null;
+
     }
 }
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
@@ -30,7 +30,7 @@ public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
             return null;
         }
 
-        return null;
+        return new DocumentInfo();
 
     }
 }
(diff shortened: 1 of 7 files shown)


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -1,12 +1,36 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.model.transaction.Transaction;
+import uk.gov.companieshouse.document.generator.accounts.service.TransactionService;
 import uk.gov.companieshouse.document.generator.interfaces.DocumentInfoService;
 import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfo;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
 
+@Component
```
BSaunders95: Can you annotate this with `@Service`, it's only a semantics change but keeps consistency with the approach we're following in the web.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -1,12 +1,36 @@
...
 public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
 
+    @Autowired
+    private TransactionService transactionService;
+
+    public static final String MODULE_NAME_SPACE = "document-generator-accounts";
```
BSaunders95: Rightly or wrongly, we're storing application namespace in the main files now -see https://github.com/companieshouse/company-accounts.api.ch.gov.uk/blob/develop/src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java#L22

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: Yep, good point, I named it `@Component` when it was under a different name, since I changed it to service I forgot to do the correct annotation :+1:

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/transaction/TransactionManager.java
```diff
@@ -0,0 +1,87 @@
...
+        return new EnvironmentReaderImpl().getMandatoryString("CHS_API_KEY");
+    }
+
+    /**
+     * Builds the private transaction link with the transaction id
+     *
+     * @return Returns the private endpoint transaction url with the transaction id
+     */
+    private static String getBaseUrl(String id) {
+        return "/private/transactions/" + id;
```
BSaunders95: Sonar tends to complain about string concatenation; use a StringBuilder here instead.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: Very true, I believe it is the correct thing to do as it is all the same in most cases, just saves someone having to change the name in multiple places. We have also done it for `document-generator-core` however since we are using maven module, it would make sense that we split it to module specific logging as you don't want a `document-generator-accounts` log to come out under `document-generator-core`. Additionally `document-generator-accounts` hasn't got a main file in regards to `SpringBoot` and `main` but this is the only entry point to this module so it made sense to put it in here.

BSaunders95: Fair point, disregard my comment!

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
BSaunders95: 👍

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -1,12 +1,36 @@
...
     @Override
     public DocumentInfo getDocumentInfo() {
-        return null;
+        LOG.info("Started getting document");
+
+        String resource = "";
+
+        Transaction transaction = transactionService.getTransaction(resource);
+        if (transaction == null) {
+            return null;
```
drosser: Should we log if the transaction is null, to indicate that something has happened? Is returning null here safe?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/transaction/TransactionManager.java
```diff
@@ -0,0 +1,91 @@
...
+        return restTemplate.exchange(getRootUri() + url, HttpMethod.GET, requestEntity, Transaction.class);
+    }
+
+    /**
+     * Get the API_URL environment variable
+     *
+     * @return Get the API_URL if set, otherwise throw an exception
+     */
+    private static String getRootUri() {
+        return new EnvironmentReaderImpl().getMandatoryString("API_URL");
```
drosser: I'm not sure this is safe. `getMandatoryString` throws a `RuntimeException` in the form of `EnvironmentVariableException` (see https://github.com/companieshouse/environment-reader-library/blob/develop/src/main/java/uk/gov/companieshouse/environment/impl/EnvironmentReaderImpl.java#L28 and https://github.com/companieshouse/environment-reader-library/blob/develop/src/main/java/uk/gov/companieshouse/environment/exception/EnvironmentVariableException.java). 

If we need a value from an environment variable, we should be getting it on startup, and failing to start the application if we can't get it. I understand the structure of this application is unique in the sense that you won't necessarily know what environment variables you'll need on startup, but that's something we probably need to look at as a problem.

That said, I feel like `API_URL` and `CHS_API_KEY` are pretty fundamental to the purpose of this API, so we should be checking these on startup, and shutting down the service if we can't get them.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: So this is part of the skeleton, it's still up in the air on what exactly I return. So I'll leave it null for the time being, however, I have missed out a logging line here to say the transaction is null, so I will add that :+1:

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/transaction/TransactionManager.java
ChrisJBurns: How would that work when the private sdk gets introduced? It will handle all of this as this temporarily is just an alternative until the private sdk is done. So how can we stop our service from starting by checking a variable that this module doesn't need to know about?

ChrisJBurns: As discussed, this will be completed by issue  [#14](https://github.com/companieshouse/document-generator/issues/14)

PR #13, opened 2018-08-22
Initial creation of DocumentGenertor in document-generator-core

Mathsh:
Initial Creation of DocumentGenerator Skeleton Class to poll kafka, deserialise message and make call to DocumentInfoService.

Updated references to @service to be @component
Added @component to AvroDerserializer and Serializers
Created a ConsumerGroupHandler class with interface
Created DocumentGeneratorTest with 1 test to check poll con confirm

resolves SFA - 421

### DIFF
--- a/document-generator-core/pom.xml
+++ b/document-generator-core/pom.xml
@@ -70,6 +70,12 @@
             <artifactId>commons-io</artifactId>
             <version>${commons-io.version}</version>
         </dependency>
+        <dependency>
+            <groupId>uk.gov.companieshouse</groupId>
+            <artifactId>document-generator-interfaces</artifactId>
+            <version>unversioned</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/document-generator-core/pom.xml
+++ b/document-generator-core/pom.xml
@@ -22,10 +22,14 @@
         
         <!-- JDK Version -->
         <jdk.version>1.8</jdk.version>
+
+        <!-- Dependancy Version -->
+        <environment-reader-library.version>1.0.0-rc1</environment-reader-library.version>
         
         <!-- JUnit Testing -->
         <junit-jupiter-engine.version>5.2.0</junit-jupiter-engine.version>
         <mockito-junit-jupiter.version>2.18.0</mockito-junit-jupiter.version>
+
         <!-- Structured logging -->
         <structured-logging.version>1.2.0-rc1</structured-logging.version>
         
@@ -37,7 +41,13 @@
     </properties>
   
     <dependencies>
-        <!--  Test  -->
+        <dependency>
+            <groupId>uk.gov.companieshouse</groupId>
+            <artifactId>environment-reader-library</artifactId>
+            <version>${environment-reader-library.version}</version>
+        </dependency>
+
+        <!-- JUnit Testing -->
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
@@ -50,19 +60,19 @@
             <version>${mockito-junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <!-- Structured logging -->
+        <dependency>
+            <groupId>uk.gov.companieshouse</groupId>
+            <artifactId>structured-logging</artifactId>
+            <version>${structured-logging.version}</version>
+        </dependency>
         
         <!-- Kafka -->
          <dependency>
             <groupId>uk.gov.companieshouse</groupId>
             <artifactId>ch-kafka</artifactId>
             <version>${ch-kafka.version}</version>
-        </dependency>
-        
-        <!-- Structured logging -->
-        <dependency>
-            <groupId>uk.gov.companieshouse</groupId>
-            <artifactId>structured-logging</artifactId>
-            <version>${structured-logging.version}</version>
         </dependency>
         
          <dependency>
@@ -74,7 +84,6 @@
             <groupId>uk.gov.companieshouse</groupId>
             <artifactId>document-generator-interfaces</artifactId>
             <version>unversioned</version>
-            <scope>compile</scope>
         </dependency>
     </dependencies>
 
--- a/document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
+++ b/document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
@@ -4,7 +4,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import uk.gov.companieshouse.document.generator.core.avro.AvroDeserializer;
 import uk.gov.companieshouse.document.generator.core.document.models.RenderSubmittedDataDocument;
 import uk.gov.companieshouse.document.generator.core.kafka.ConsumerGroupHandler;
-import uk.gov.companieshouse.document.generator.core.kafka.configurationhelper.KafkaConfigHelper;
 import uk.gov.companieshouse.document.generator.interfaces.DocumentInfoService;
 import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfo;
 import uk.gov.companieshouse.kafka.consumer.CHKafkaConsumerGroup;
@@ -21,22 +20,15 @@ public class DocumentGenerator implements Runnable {
     
     private CHKafkaConsumerGroup documentGeneratorConsumerGroup;
 
-    private KafkaConfigHelper kafkaConfigHelper;
-
     private DocumentInfoService documentInfoService;
-
-    private ConsumerGroupHandler consumerGroupHandler;
 
     private AvroDeserializer<RenderSubmittedDataDocument> avroDeserializer;
 
     @Autowired
-    public DocumentGenerator(KafkaConfigHelper kafkaConfigHelper, DocumentInfoService documentInfoService,
-                             ConsumerGroupHandler consumerGroupHandler,
+    public DocumentGenerator(DocumentInfoService documentInfoService, ConsumerGroupHandler consumerGroupHandler,
                              AvroDeserializer<RenderSubmittedDataDocument> avroDeserializer) {
 
-        this.kafkaConfigHelper = kafkaConfigHelper;
         this.documentInfoService = documentInfoService;
-        this.consumerGroupHandler = consumerGroupHandler;
         this.avroDeserializer = avroDeserializer;
 
         documentGeneratorConsumerGroup =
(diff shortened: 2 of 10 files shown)


document-generator-core/pom.xml
```diff
@@ -70,6 +70,12 @@
             <artifactId>commons-io</artifactId>
             <version>${commons-io.version}</version>
         </dependency>
+        <dependency>
+            <groupId>uk.gov.companieshouse</groupId>
+            <artifactId>document-generator-interfaces</artifactId>
+            <version>unversioned</version>
+            <scope>compile</scope>
```
ChrisJBurns: can get rid of that, as its the default scope

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.companieshouse.document.generator.interfaces.DocumentInfoService;
+import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfo;
+import uk.gov.companieshouse.kafka.consumer.CHKafkaConsumerGroup;
+import uk.gov.companieshouse.kafka.message.Message;
+
+import java.util.Arrays;
+
+public class DocumentGenerator implements Runnable {
+
+    private static final String[] CONSUMER_TOPICS = new String[]{
```
ChrisJBurns: Any reason we have a String array here? and not a list? That would save us having to convert to a list when we use it?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+public class DocumentGenerator implements Runnable {
+
+    private static final List<String> CONSUMER_TOPICS = Arrays.asList("render-submitted-data-document");
+
+    private static final String GROUP_NAME = "document-generator";
+    
+    private CHKafkaConsumerGroup documentGeneratorConsumerGroup;
```
drosser: This belongs to the `ConsumerGroupHandler`, but we still hold responsibility here. You could remove this variable and just get it by calling `consumerGroupHandler.getConsumerGroup()`

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+
+        documentGeneratorConsumerGroup =
+                consumerGroupHandler.getConsumerGroup(CONSUMER_TOPICS, GROUP_NAME);
+    }
+
+    @Override
+    public void run() {
+
+        try {
+            pollAndGenerate(documentGeneratorConsumerGroup);
```
drosser: `documentGeneratorConsumerGroup` is a member variable - it should either be removed (see previous comment), or it doesn't need to be passed in here.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+
+        for (Message message : documentGeneratorConsumerGroup.consume()) {
+
+            RenderSubmittedDataDocument renderSubmittedDataDocument;
+
+            try {
+                renderSubmittedDataDocument = avroDeserializer
+                        .deserialize(message, RenderSubmittedDataDocument.getClassSchema());
+
+                DocumentInfo documentInfo = documentInfoService.getDocumentInfo();
```
drosser: This is never used

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+
+                documentGeneratorConsumerGroup.commit();
+            } catch (Exception e) {
+                documentGeneratorConsumerGroup.commit();
```
drosser: If we're committing whether we succeed or fail, we could move this (and the previous line that does the same thing) outside of the `try/catch` and call it once

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
Mathsh: because it is a skeleton DocumentGenerator. It will be used when a future task for document-generator-core is implemented

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
Mathsh: it is done like this with regards to getting an error we cannot recover from, it will hold the queue up unless we commit.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+                consumerGroupHandler.getConsumerGroup(CONSUMER_TOPICS, GROUP_NAME);
+    }
+
+    @Override
+    public void run() {
+
+        try {
+            pollAndGenerate();
+        } catch (Exception e) {
+
```
jsimons1989: Exceptions must not be ignored. There should not be an empty catch block.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
jsimons1989: If a different developer picked up the next task, how would they know what your intention was with the deserialiser and documentInfo? Would logging or a TODO linked to the next task be beneficial?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+            RenderSubmittedDataDocument renderSubmittedDataDocument;
+
+            try {
+                renderSubmittedDataDocument = avroDeserializer
+                        .deserialize(message, RenderSubmittedDataDocument.getClassSchema());
+
+                DocumentInfo documentInfo = documentInfoService.getDocumentInfo();
+
+                documentGeneratorConsumerGroup.commit();
+            } catch (Exception e) {
```
jsimons1989: This needs logging to be able to diagnose errors. Can you catch the specific errors to write logs appropriate to the error which has occurred please?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/kafka/impl/ConsumerGroupHandlerImpl.java
```diff
@@ -0,0 +1,29 @@
...
+     * get the consumer group
+     *
+     * @param cosumerTopics
+     * @param groupName
+     * @return
+     */
+    @Override
+    public CHKafkaConsumerGroup getConsumerGroup(List<String> cosumerTopics, String groupName) {
+        return new CHKafkaConsumerGroup(kafkaConfigHelper
+                .configureKafkaConsumer(cosumerTopics, groupName));
```
jsimons1989: Why is an interface needed for this helper class? We won't have more than 1 implementation for getting a consumer group.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/kafka/impl/ConsumerGroupHandlerImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * get the consumer group
+     *
+     * @param cosumerTopics
+     * @param groupName
+     * @return
+     */
+    @Override
+    public CHKafkaConsumerGroup getConsumerGroup(List<String> cosumerTopics, String groupName) {
```
jsimons1989: Typo in "cosumerTopics"

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/document/DocumentGeneratorTest.java
```diff
@@ -0,0 +1,138 @@
...
+    @BeforeEach
+    public void setUp() {
+
+        createTestMessageList();
+        createTestRenderSubmittedDataDocument();
+    }
+
+    @Test
+    @DisplayName("test the polling of Kafka and the consumer group commit")
+    public void testPollAndConsumerGroupCommit() throws IOException {
```
jsimons1989: You're missing a test for verifying a commit when an exception is caught

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/kafka/impl/ConsumerGroupHandlerImpl.java
Mathsh: good practice to code to interface to decouple from there implementations, one I follow. granted there are arguments against it. And yes it was extracted to its own class to allowing mocking as we are moving away from powerMock which caused us issues.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
Mathsh: I have a task to implement all logging, which will be using the structured logging development work that Guru undertook.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
Mathsh: as below

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
jsimons1989: Code in Develop should be releasable. (see [GitFlow at CH](https://companieshouse.atlassian.net/wiki/spaces/DEVOPS/pages/78151945/GitFlow+at+CH#GitFlowatCH-developDevelopBranch)) which means all exceptions should be handled and logged appropriately (see [Java review standards](https://github.com/companieshouse/styleguides/blob/master/java_review.md)). Other developers could try to integrate with your code or take over one of your tasks at any time and will be frustrated if they have problems and nothing is getting logged with exceptions ignored.

Until Guru's code is ready can you add an import for the structured-logger please.

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/document/DocumentGeneratorTest.java
Mathsh: added test to check

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/document/DocumentGenerator.java
Mathsh: added Log.error in catch

PR #16, opened 2018-08-31
Added Abridged account logic to determine if accounts is abridged

ChrisJBurns:
- gets and checks resource in transaction is abridged accounts and if so, it does some abridged logic (to be implemented in next PR - currently it just returns an empty `DocumentInfo` object to split functionality across smaller PRs.
- added unit tests to cover new scenarios of no transaction and no account resource

Implements: SFA-567

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
@@ -1,7 +1,9 @@
 package uk.gov.companieshouse.document.generator.accounts;
 
+import java.util.Optional;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import uk.gov.companieshouse.api.model.transaction.Resource;
 import uk.gov.companieshouse.api.model.transaction.Transaction;
 import uk.gov.companieshouse.document.generator.accounts.service.TransactionService;
 import uk.gov.companieshouse.document.generator.interfaces.DocumentInfoService;
@@ -24,6 +26,7 @@ public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
         LOG.info("Started getting document");
 
         String resource = "";
+        String resourceId = "";
 
         Transaction transaction = transactionService.getTransaction(resource);
         if (transaction == null) {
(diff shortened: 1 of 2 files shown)


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -24,14 +26,42 @@ public DocumentInfo getDocumentInfo() {
...
         Transaction transaction = transactionService.getTransaction(resource);
         if (transaction == null) {
             LOG.error("transaction not found");
             return null;
         }
 
-        return new DocumentInfo();
+        String accountLink = Optional.of(transaction)
+                .map(Transaction::getResources)
+                .map(e -> e.get(resourceId))
```
jterrych: would change from e to resources, makes it more readable.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -24,14 +26,42 @@ public DocumentInfo getDocumentInfo() {
...
+                .map(Resource::getLinks)
+                .map(e -> e.get("resource"))
```
jterrych: would change from e to links, as above

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -24,14 +26,42 @@ public DocumentInfo getDocumentInfo() {
...
+                .orElseGet(() -> {
+                    LOG.error("Unable to find account resource link in transaction under:" + resourceId);
+                    return Optional.empty().toString();
+                });
 
+        // when abridged has been migrated to use the company-accounts api, the code for the
+        // company accounts should work for abridged, resulting in this abridged specific code
+        // qualifying for removal
+        if (isAbridged(accountLink)) {
```
jterrych: Only a small suggestion - could add this to the optional which itself could be returned.
```
return Optional.of(transaction)
              ...
              ...
              .filter(this::isAbridged)
              .map(accountLinks -> new DocumentInfo())
              .orElse(null);
```

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: good point, changed

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: good point, changed

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: Yeah good shout :+1:

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -24,14 +26,35 @@ public DocumentInfo getDocumentInfo() {
...
     }
+
+    /**
+     * determines whether it is an abridged link as /accounts/ only exists within the abridged
+     * implementation
+     * @param accountLink - account link
+     * @return true if abridged, false if not
+     */
+    private boolean isAbridged(String accountLink) {
+        return accountLink.contains("/accounts/");
```
BSaunders95: Should we be using a regex matcher here to more confidently ensure it is an abridged account? May not be needed, just reduces the chance of following the wrong 'flow' in the case of bad data?

ChrisJBurns: added. I could have put it into a variable but to be honest its only going to be used in one place so we shouldn't have to worry.

PR #17, opened 2018-09-04
Adds accounts retrieval implementation

ChrisJBurns:
Adds the accounts retrieval implementation for abridged. Includes:
- AccountType enum for document generator
- temporary private api sdk implementation for retrieval of accounts data
- AccountsService that interacts with the private-sdk:
    - adds the AccountsService interfact
    - adds the AccountsServiceImpl implementation
    - adds base test file ready for when the temporary private sdk solution is replaced by permanent one
- accounts retrieval logic for abridged
    - adds accounts retrieval logic for the abridged implementation
    - modifies test to deal with new scenarios of null accounts data from service and incorrect AccountType

Implements: SFA-567


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    @Override
+    public Accounts getAccounts(String resource) {
+        LOG.info("Getting accounts data from abridged-accounts-api");
+
+        ResponseEntity<Accounts> accounts = AccountsManager.getAccounts(resource);
+
+        if (accounts.getStatusCode() != HttpStatus.OK) {
+            LOG.error("Failed to retrieve data from API: " + resource);
+            return null;
```
BSaunders95: Is this an exceptional case? If so it may be more appropriate to throw an exception here and catch it higher up? This would save excessive null checking.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+public class AccountsServiceImpl implements AccountsService {
+
+    private static final Logger LOG = LoggerFactory.getLogger("document-generator-accounts");
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Accounts getAccounts(String resource) {
+        LOG.info("Getting accounts data from abridged-accounts-api");
```
BSaunders95: Is it worth including the `resource` in the log so it's more meaningful to support?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
ChrisJBurns: Yeah, exception handling, logging and the abridged implementation will be done next. I'm just putting the skeleton functionality in initially and will then apply the correct exception handling and logging across the board. Currently I am putting in null checks just to keep the files down to a minimum.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
ChrisJBurns: As mentioned in my other comment, logging will be fine-tuned in a separate PR :)

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
BSaunders95: 👍

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class AccountsManager {
+
+    /** represents the Authorization header name in the request */
+    private static final String AUTHORIZATION_HEADER = "Authorization";
+
+    private static final RestTemplate restTemplate = createRestTemplate();
+
+    /**
+     * Try get transaction if exists
```
BSaunders95: Do you mean 'Try to get accounts if they exist' or similar?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @param link - self link for the accounts object
+     * @return transaction object along with the status or not found status.
```
BSaunders95: Again, this should be relevant to accounts, not transactions

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -0,0 +1,73 @@
...
+        return restTemplate.exchange(getRootUri() + url, HttpMethod.GET, requestEntity, Accounts.class);
+    }
+
+    /**
+     * Get the root uri from the properties file
+     *
+     * @return Get the root uri if set, otherwise throw an exception
+     */
+    private static String getRootUri() {
+        return new EnvironmentReaderImpl().getMandatoryString("API_URL");
```
BSaunders95: If a static `EnvironmentReaderImpl` is instantiated at the class level, I believe we can have static Strings on the class to hold these details; that way, Strings are initialised once and we don't create a new reader instance in every method call. Something like:

```
public class Class {

   private static String API_URL = READER.getMandatoryString("API_URL");

}
```

The Strings might even be able to be made final - not sure about that without checking though.

The added benefit of this is that any `MissingEnvironmentVariableException`'s will be thrown on class instantiation, as opposed to when the existing method is called.

BSaunders95: (I know this will be stripped out, but it can't hurt to do it in an efficient way for the time being.)

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImplTest.java
```diff
@@ -30,16 +32,13 @@
...
     void testUnsuccessfulGetDocumentInfoNoAccountsResourceInTransaction() {
-        Map<String, Resource> resources = new HashMap<>();
-        resources.put("error", createResource());
-
-        Transaction transaction = new Transaction();
-        transaction.setResources(resources);
-        when(transactionService.getTransaction(anyString())).thenReturn(transaction);
-
+        when(transactionService.getTransaction(anyString())).thenReturn(createTransaction());
         assertNull(accountsDocumentInfoService.getDocumentInfo());
```
BSaunders95: Should there also be an accountService call here too? Which, given the context of the test, would return null I would assume?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/AccountType.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public String getResourceKey() {
+        return resourceKey;
+    }
+
+    public static AccountType getAccountType(String key) {
+        return Arrays.stream(AccountType.values())
+                .filter(e -> e.getResourceKey().equalsIgnoreCase(key))
+                .findFirst()
+                .orElse(null);
```
BSaunders95: I would believe this to be exceptional, so may be worth using `.orElseThrow()` and throw a checked exception here. But I understand exception handling will be done in later PR's.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -43,11 +49,48 @@ public DocumentInfo getDocumentInfo() {
                 // company accounts should work for abridged, resulting in this abridged specific code
                 // qualifying for removal
                 .filter(this::isAbridged)
-                .map(accountsLinks -> new DocumentInfo())
+                .map(accountsLinks -> getAbridgedAccountData(resourceId))
```
jsimons1989: How is this scalable to support other account types?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -43,11 +49,48 @@ public DocumentInfo getDocumentInfo() {
...
+        Accounts accounts = accountsService.getAccounts(resourceId);
+
+        if (accounts == null) {
+            return null;
+        }
+
+        AccountType accountsType = getAccountType(accounts);
+        if (accountsType == null) {
+            return null;
+        }
```
jsimons1989: Why is this code in an abridged specific method? There's nothing specific to abridged in here.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.web.client.RestTemplate;
+import uk.gov.companieshouse.api.model.accounts.Accounts;
+import uk.gov.companieshouse.environment.impl.EnvironmentReaderImpl;
+
+public class AccountsManager {
+
+    /** represents the Authorization header name in the request */
+    private static final String AUTHORIZATION_HEADER = "Authorization";
+
+    private static final RestTemplate restTemplate = createRestTemplate();
```
jsimons1989: How are methods using the RestTemplate going to be tested if it's statically created on class load?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/AccountsService.java
```diff
@@ -0,0 +1,20 @@
...
+
+public interface AccountsService {
+
+    /**
+     * Call the api sdk to get accounts data of resource passed.
+     *
+     * __NOTE:__: It is important to note, currently it will call an internal accounts manager
+     * within this project that calls the abridged api through ERIC as the private sdk hasn't
+     * been built yet. When the private SDK has been built, we should be able to simply delete the
+     * accounts package and use the SDK in the implementation of this method.
```
jsimons1989: Can this be linked to the task to implement the private SDK please. This will allow developers reading this to be able to easily check if the dependent work is completed.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/AccountsService.java
```diff
@@ -0,0 +1,20 @@
...
+     *
+     * @return transaction
```
jsimons1989: This is returning an account not a transaction.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
jsimons1989: Logging should contain all relevant information. This could be missed in a future PR.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+        ResponseEntity<Accounts> accounts = AccountsManager.getAccounts(resource);
+
+        if (accounts.getStatusCode() != HttpStatus.OK) {
+            LOG.error("Failed to retrieve data from API: " + resource);
+            return null;
+        }
+        LOG.trace("Accounts data retrieved successfully");
```
jsimons1989: Is this log going to benefit support or cause noise in the logs (it has no reference to which account was retrieved and cannot be linked to the request)?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Service;
+import uk.gov.companieshouse.api.model.accounts.Accounts;
+import uk.gov.companieshouse.document.generator.accounts.data.accounts.AccountsManager;
+import uk.gov.companieshouse.document.generator.accounts.service.AccountsService;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
+
+@Service
+public class AccountsServiceImpl implements AccountsService {
```
jsimons1989: Shouldn't this be called AbridgedAccountsServiceImpl? The JavaDoc on the method which calls this class says "Gets the abridged account data from abridged specific services". The data in each account type is different and the endpoints could be different, they should be in different implementations?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -43,11 +49,48 @@ public DocumentInfo getDocumentInfo() {
...
+    private AccountType getAccountType(Accounts accountsData) {
+        return accountsData.getLinks().keySet()
+                .stream()
+                .filter(e -> !e.equalsIgnoreCase("self"))
+                .map(AccountType::getAccountType)
+                .findFirst()
+                .orElseGet(() -> {
+                    LOG.error("Unable to find account type in account data: " + accountsData.getId());
+                    return null;
+                });
```
paulforsyth: if this is an error why are you returning null?

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImplTest.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.companieshouse.document.generator.accounts.service.impl;
+
+public class AccountsServiceImplTest {
+
```
jsimons1989: This is missing unit tests.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
ChrisJBurns: It won't be missed purely because I've already done the work and I am simply copy and pasting small chunk at a time just to keep the PRs small.

Regarding the second point, the reason why `abridged-api` is mentioned is because I want there to be a clear separation between the way we use the `company-accounts.api` and the existing `abridged-api`. We already know that it will eventually get migrated to the new api, but for the time being I have made a clear distinction because it will make it much easier to take out in future and makes it easier to debug in regards to knowing where the error comes from.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
ChrisJBurns: Its worth noting that this will be replaced by the `private-sdk' soon so we shouldn't get too tied down to making this perfect - however you raise a good point so I will change.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: It's pretty scalable as there is only two directions it can go. Either abridged or company-accounts. That covers all account types. The company-accounts direction will exist in the `orElse`

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
ChrisJBurns: This is the alternative `private-sdk` implementation. Currently we cannot JUnit test it. This is the same in the `company-accounts.api` also.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
ChrisJBurns: I believe it is handy to know that it has been retrieved successfully. Additionally, it isn't retrieving a certain type of accounts - just the base account resource.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
ChrisJBurns: Sorry, this is the correct name. The JD is also correct. I guess this is the downside of copy and pasting an already written solution just to make PRs smaller. They are both correct, however the next PR I will insert the abridged implementation. Which is where it will all make sense.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: What do you suggest returning?

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImplTest.java
ChrisJBurns: This is just the skeleton of the file

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImplTest.java
```diff
@@ -30,16 +32,13 @@
     @Mock
     private TransactionService transactionService;
 
+    @Mock
+    private AccountsService accountsService;
+
```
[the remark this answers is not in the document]
ChrisJBurns: `testUnsuccessfulGetDocumentInfoFailedTransactionRetrieval`

ChrisJBurns:
I have altered the PR name just so there is no mention of abridged to make the context more clear as the abridged implementation will come in the next PR

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
jsimons1989: What about supporting small full, micro, dormant, etc?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
ChrisJBurns: yes i do :)

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
ChrisJBurns: :+1:

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    private static RestTemplate createRestTemplate() {
+        HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
+        return new RestTemplate(requestFactory);
+    }
+
+
+
+    /**
+     * GET the transaction resource
```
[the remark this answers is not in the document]
ChrisJBurns: :+1:

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
jsimons1989: Sorry if I was unclear. By "no reference to which account" I meant no identifier (e.g. account ID, account type, transaction ID) .

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/AccountType.java
ChrisJBurns: Yeah I only followed what is done in the existing `document-generator-accounts`. However I would say that it isn't really an exceptional case, my reasoning for that would be typically you wouldn't throw an exception if something doesn't exist in an enum or collection as that's not really an error - rather it just .. not being there.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: That can be done in 1 method as their data is all coming from 1 api. The only reason for it splitting up is because the `abridged` data comes from a separate api. Additionally, in future, if there is more than 1 direction (which after the abridged migration - we haven't recognised one), you can add that here using multiple filters and maps.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
ChrisJBurns: Ahhh I see, I am going to tweak the logging to make it more robust in following PRs but just to get something in I have added the resource.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: Actually this isn't an error in my opinion so I will change it to log it as debug instead :+1:

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImplTest.java
ChrisJBurns: Nope, so that is why it is empty. The ones that has an accounts resource are those that have an accounts resource valid link. The resource that is empty is not a valid resource.

ChrisJBurns:
@ch-tstrickland This PR is outdated now as most of this functionality is in develop. This was too big of a PR for people to review so it was done in smaller ones. I can effectively close this now.

PR #19, opened 2018-09-05
document-generator swagger specs for api

Mathsh:
Created a new swagger spec for document-generator as document-generator-core will now become an API. 

New model also created to spec the request and response model the api will use. 
Currently as this is a private spec it is situated within document-generator-core under the folder swagger-2.0. Tested and run through dappperdox.

Resolves SFA - 664

### DIFF
--- a/document-generator-core/swagger-2.0/models/document-generator.json
+++ b/document-generator-core/swagger-2.0/models/document-generator.json
@@ -52,7 +52,7 @@
         },
         "size": {
           "type": "string",
-          "description": "size of the document stored"
+          "description": "Long value to denote the size of the document stored"
         },
         "description": {
           "type": "string",
@@ -64,10 +64,9 @@
         },
         "description_values": {
           "type": "object",
-          "description": "List of resources contained within this transaction.",
+          "description": "a map of key/value pairs used to complete the looked up description identifier enum to turn it into a full description",
           "additionalProperties":{
-            "type": "string",
-            "description": "Map of resource 'identifier' to resource object"
+            "type": "string"
           }
         }
       }
--- a/document-generator-core/swagger-2.0/models/document-generator.json
+++ b/document-generator-core/swagger-2.0/models/document-generator.json
@@ -3,56 +3,49 @@
     "documentRequest": {
       "title": "DocumentRequest",
       "required": [
-        "id",
-        "resource",
+        "resource_uri",
         "resource_id",
-        "content_type",
-        "document_type",
-        "user_id"
+        "mime_type"
       ],
       "properties": {
-        "id": {
+        "resource_uri": {
           "type": "string",
-          "description": "id for the document request"
-        },
-        "resource": {
-          "type": "string",
-          "description": "link to the resource"
+          "description": "link to the resource (currently it is the link to the parent transaction)"
         },
         "resource_id": {
           "type": "string",
-          "description": "id of the resource"
+          "description": "id of the resource (currently denotes which resource within the transaction)"
         },
-        "content_type": {
+        "mime_type": {
           "type": "string",
-          "description": "the content type of the template to be converted"
+          "description": "mime type that the document should be rendered as"
         },
         "document_type": {
           "type": "string",
-          "description": "type of document to be rendered"
-        },
-        "user_id": {
-          "type": "string",
-          "description": "id of the user requesting the document"
+          "description": "indicates the document or report type to be generated from the resource data. Currently not implemented"
         }
+
       }
     },
     "documentResponse": {
       "title": "DocumentResponse",
       "required": [
-        "location",
+        "links",
         "size",
         "description",
         "description_identifier",
       ],
       "properties": {
-        "location": {
-          "type": "string",
-          "description": "location that the document has been stored"
+        "links": {
+          "type": "object",
+          "description": "links to related resources",
+          "items": {
+            "$ref": "http://localhost:3123/swagger-2.0/models/document-generator.json#/definitions/documentLinks"
+          }
         },
         "size": {
           "type": "string",
-          "description": "Long value to denote the size of the document stored"
+          "description": "value to denote the size of the document stored in bytes"
         },
         "description": {
           "type": "string",
@@ -68,6 +61,18 @@
           "additionalProperties":{
             "type": "string"
           }
+        }
+      }
+    },
+    "documentLinks": {
+      "title": "DocumentLinks",
+      "required": [
+        "location"
+      ],
+      "properties": {
+        "location": {
+          "type": "string",
+          "description": "location that the document has been stored"
         }
       }
     }
--- a/document-generator-core/swagger-2.0/models/document-generator.json
+++ b/document-generator-core/swagger-2.0/models/document-generator.json
@@ -1,7 +1,7 @@
 {
   "definitions": {
-    "documentGeneratorRequest": {
-      "title": "DocumentGeneratorRequest",
+    "documentRequest": {
+      "title": "DocumentRequest",
       "required": [
         "id",
         "resource",
@@ -37,8 +37,8 @@
         }
       }
     },
-    "documentGeneratorResponse": {
-      "title": "DocumentGeneratorResponse",
+    "documentResponse": {
+      "title": "DocumentResponse",
       "required": [
         "location",
         "size",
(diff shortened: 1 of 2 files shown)


document-generator-core/swagger-2.0/models/document-generator.json
```diff
@@ -0,0 +1,76 @@
...
+        "resource",
+        "resource_id",
+        "content_type",
+        "document_type",
+        "user_id"
+      ],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "id for the document request"
```
macreman: what does the id actually denote? what is it for?

document-generator-core/swagger-2.0/spec/document-generator.json
```diff
@@ -0,0 +1,57 @@
...
+  "schemes": [
+    "http"
+  ],
+  "tags": [
+    {
+    "name": "document-generator"
+    }
+  ],
+  "paths": {
+    "/document-generate": {
```
macreman: should be `private` and the URI space should be built from the resource type of `documents` followed by the action to `generate` so should be:

`/private/documents/generate`

document-generator-core/swagger-2.0/models/document-generator.json
```diff
@@ -0,0 +1,76 @@
+{
+  "definitions": {
+    "documentGeneratorRequest": {
```
macreman: name should be `documentRequest` to describe the data we're sending in, the title below should then be `DocumentRequest`

document-generator-core/swagger-2.0/models/document-generator.json
```diff
@@ -0,0 +1,76 @@
...
+          "type": "string",
+          "description": "type of document to be rendered"
+        },
+        "user_id": {
+          "type": "string",
+          "description": "id of the user requesting the document"
+        }
+      }
+    },
+    "documentGeneratorResponse": {
```
macreman: name should be `documentResponse` to describe the data we're sending in, the title below should then be `DocumentResponse`

document-generator-core/swagger-2.0/models/document-generator.json
```diff
@@ -0,0 +1,76 @@
...
+        },
+        "resource": {
+          "type": "string",
+          "description": "link to the resource"
+        },
+        "resource_id": {
+          "type": "string",
+          "description": "id of the resource"
+        },
+        "content_type": {
```
macreman: `content_type` in the body is confusing, this denotes the format that the document should be generated in so we should use `mime_type` and enumerate the options we support which will currently be `application/pdf` and `text/html` (for html5 and ixbrl documents)

document-generator-core/swagger-2.0/models/document-generator.json
```diff
@@ -0,0 +1,76 @@
...
+          "type": "string",
+          "description": "the content type of the template to be converted"
+        },
+        "document_type": {
```
macreman: `document_type` in the body is confusing, this denotes the format that the document should be generated in so we should use `mime_type` and enumerate the options we support which will currently be `application/pdf` and `text/html` (for html5 and ixbrl documents)

document-generator-core/swagger-2.0/models/document-generator.json
macreman: this should not be in the kafka message either

document-generator-core/swagger-2.0/models/document-generator.json
```diff
@@ -0,0 +1,76 @@
...
+        "description_identifier": {
+          "type": "string",
+          "description": "Identifier for the document description"
+        },
+        "description_values": {
+          "type": "object",
+          "description": "List of resources contained within this transaction.",
+          "additionalProperties":{
+            "type": "string",
+            "description": "Map of resource 'identifier' to resource object"
```
macreman: again, this doesn't sound quite right here, it is `a map of key/value pairs used to complete the looked up description identifier enum to turn it into a full description`

this description will no doubt be tweaked for readability in future :)

document-generator-core/swagger-2.0/spec/document-generator.json
```diff
@@ -0,0 +1,57 @@
...
+      "post": {
+        "x-operationName": "generate",
+        "summary": "Generation of documents",
+        "description": "Generate a document using the request to obtain the document data and send to render service",
```
macreman: `Takes a URI of a resource and synchronously generates a document of the requested mime type`

document-generator-core/swagger-2.0/spec/document-generator.json
```diff
@@ -0,0 +1,57 @@
...
+          "document-generator"
+        ],
+        "consumes": [
+          "text/json"
+        ],
+        "parameters": [
+          {
+            "name": "data",
+            "in": "body",
+            "description": "required parameters to generate a document",
```
macreman: `A DocumentRequest resource`

document-generator-core/swagger-2.0/spec/document-generator.json
```diff
@@ -0,0 +1,57 @@
...
+        ],
+        "responses": {
+          "201": {
+            "description": "The document has been created",
+            "schema": {
+              "$ref": "http://localhost:3123/swagger-2.0/models/document-generator.json#/definitions/documentGeneratorResponse"
+            }
+          },
+          "400": {
+            "description": "Unable to handle given request"
```
macreman: why? what conditions would result in this error status? expliciti examples can be seen in responses here http://ch-jenkins-rh1.internal.ch:3123/manipulate-company-data/reference/transactions/update

document-generator-core/swagger-2.0/spec/document-generator.json
```diff
@@ -0,0 +1,57 @@
...
+          },
+          "401": {
+            "description": "Not authorized to add the resource "
```
macreman: doesn't sound right here

Mathsh: what doesn't sound right?

document-generator-core/swagger-2.0/spec/document-generator.json
Mathsh: I was told to return 400 which as far As I was aware had been discussed. 400, 401 to be returned. If this needs to change, it can be done. But I think a clear outline of what should be returned needs to be agreed

document-generator-core/swagger-2.0/models/document-generator.json
Mathsh: This was formed after a discussion with Tech Chris

document-generator-core/swagger-2.0/models/document-generator.json
```diff
@@ -0,0 +1,76 @@
...
+          "type": "string",
+          "description": "description of the document produced"
+        },
+        "description_identifier": {
+          "type": "string",
+          "description": "Identifier for the document description"
+        },
+        "description_values": {
+          "type": "object",
+          "description": "List of resources contained within this transaction.",
```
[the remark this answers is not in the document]
Mathsh: as below

document-generator-core/swagger-2.0/spec/document-generator.json
macreman: the description `Not authorized to add the resource` - we're not adding a resource

document-generator-core/swagger-2.0/spec/document-generator.json
macreman: Rewording the initial comment - the description should inform the reader the conditions that lead to a 400 response

document-generator-core/swagger-2.0/models/document-generator.json
macreman: the description is copied directly from the transactions api spec so is dealing with resources, which are specific to transaction data, this description does not make sense in this context.

The structure of this map is fine but the description does not match the document-generator usage of this structure

document-generator-core/swagger-2.0/models/document-generator.json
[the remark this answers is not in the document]
macreman: again, the description_values object is fine but the wording in this description i.e. `List of resources contained within this transaction.` does not fit the document-generator context - there are no resources and this is not a transaction

document-generator-core/swagger-2.0/models/document-generator.json
Mathsh: id, resource, resource_id, content_type, document_type and user_id are all obtained when deserialising the message from Kafka.

PR #20, opened 2018-09-06
Creation of skeletal Rest Controller in document-generator-core

Mathsh:
Created skeletal Rest controller in document-generator-core
Created request and response model for controller
Added Junit test to test request binding errors and Null response error

Currently response will always return null as service not yet created.

Resolves SFA - 664


document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/models/DocumentGeneratorRequest.java
```diff
@@ -0,0 +1,80 @@
+package uk.gov.companieshouse.document.generator.core.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotNull;
+
+public class DocumentGeneratorRequest {
```
ChrisJBurns: i would add the `toString` method here as its good practice to include them in POJOs

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/models/DocumentGeneratorResponse.java
```diff
@@ -0,0 +1,63 @@
+package uk.gov.companieshouse.document.generator.core.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Map;
+
+public class DocumentGeneratorResponse {
```
ChrisJBurns: same with `toString()` here?

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Mock
+    private BindingResult mockBindingResult;
+
+    @InjectMocks
+    private DocumentGeneratorController documentGeneratorController;
+
+    @Test
+    @DisplayName("Tests if Internal Server Error return as document generator response is null")
+    public void errorReturnAsNullResponse() {
```
ChrisJBurns: dont need public access modifier anymore with JUnit5

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
```diff
@@ -0,0 +1,59 @@
...
+        request.setUserId("user_id");
+
+        ResponseEntity responseEntity = documentGeneratorController.generateDocument(request, mockBindingResult);
+
+        assertEquals(HttpStatus.INTERNAL_SERVER_ERROR, responseEntity.getStatusCode());
+    }
+
+    @Test
+    @DisplayName("Tests if Bad Request error return when binding error found")
+    public void errorReturnAsIncompleteRequest() {
```
ChrisJBurns: dont need public access modifier anymore with JUnit5

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @InjectMocks
+    private DocumentGeneratorController documentGeneratorController;
+
+    @Test
+    @DisplayName("Tests if Internal Server Error return as document generator response is null")
+    public void errorReturnAsNullResponse() {
+
+        DocumentGeneratorRequest request = new DocumentGeneratorRequest();
+        request.setContentType("content_type");
```
ChrisJBurns: if this is going to be reused in future tests, could be worth creating a method that creates this `DocumentGeneratorRequest` object\?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
```diff
@@ -0,0 +1,55 @@
...
+        } catch (Exception e) {
+            LOG.error(e);
+        }
+
+        if(response == null) {
+            LOG.error("no data has been returned");
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+
+        if(response.getLocation() != null) {
```
ChrisJBurns: Is this tested? I feel that it is possibly missed because of the `response` always been null.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.validation.Valid;
+
+import static uk.gov.companieshouse.document.generator.core.DocumentGeneratorApplication.APPLICATION_NAME_SPACE;
+
+@RestController
+public class DocumentGeneratorController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(APPLICATION_NAME_SPACE);
+
+    @PostMapping("/document-generate")
```
BSaunders95: Can you please move this to a `@RequestMapping()` at the class level to make it consistent with the web and api. That way, this line can just become `@PostMapping` which makes the class a little cleaner

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
```diff
@@ -0,0 +1,55 @@
...
+        DocumentGeneratorResponse response = new DocumentGeneratorResponse();
+
+        if (result.hasErrors()) {
+            LOG.error("error in request body");
+            return new ResponseEntity<>(result.getAllErrors(), HttpStatus.BAD_REQUEST);
+        }
+
+        try {
+            response = null;
+        } catch (Exception e) {
```
BSaunders95: What are we catching here?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/models/DocumentGeneratorRequest.java
Mathsh: done

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/models/DocumentGeneratorResponse.java
Mathsh: done

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
Mathsh: makes no difference having it there or not. Prefer to physically see it

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
Mathsh: makes no difference having it there or not. Prefer to physically see it

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
Mathsh: If it is reused it will be refactored into its own method.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
Mathsh: As stated in the pr, this is a skeleton controller, I know I will catch an exception here when the service is implemented so I have added the Exception to the catch to ensure it is not forgotten.

BSaunders95: Is it worth adding a basic service interface so that it can be mocked here and test coverage improved?

Mathsh: it is the next stage I am doing, is to implement the service. So in the next process test coverage will be improved.

ChrisJBurns:
:+1:

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
BSaunders95: In my opinion we shouldn't be try-catching where it's not appropriate. This could be changed either by adding an interface which throws an expected exception - which would enable more complete testing of the controller - or by refactoring the controller to just set the `response` object to `null` without the try-catch. If the latter is chosen, it could be accompanied by a TODO with a relevant SFA number.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/models/DocumentGeneratorRequest.java
BSaunders95: We could use GSON or Jackson in the `toString` method which would remove any risk of a variable name mistype. Not needed for this PR, just food for thought.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/models/DocumentGeneratorRequest.java
```diff
@@ -0,0 +1,92 @@
...
+        return userId;
+    }
+
+    public void setUserId(String userId) {
+        this.userId = userId;
+    }
+
+    @Override
+    public String toString() {
+        try {
```
BSaunders95: I believe this can be performed as:
```
public String toString() {
    Gson gson = new Gson();
    return gson.toJson(this);
}
```
Which saves the need of a try-catch

Mathsh: yes it does, but it's a google library. We currently have Jackson. If the leads agree we can use it i'll change it.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
```diff
@@ -0,0 +1,62 @@
...
+
+        if(response == null) {
+            LOG.error("no data has been returned");
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+
+        if(response.getLinks() != null) {
+            return new ResponseEntity<>(response, HttpStatus.CREATED);
+        } else {
+            LOG.error("Failed to generate the document");
```
BSaunders95: Is it worth logging some request details in here to aid support in troubleshooting?

Mathsh: There's a separate story to flesh out all the logging to include the new structured logging.

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @Test
+    @DisplayName("Test if Created returned if response not null and location present")
+    public void createdReturned() {
+
+        DocumentRequest request = setDocumentgeneratorRequest();
+
+        DocumentResponse response = new DocumentResponse();
+        response.setDescription("description");
```
BSaunders95: These variables are used more than once and could be made `static final` on the class

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorControllerTest.java
```diff
@@ -0,0 +1,156 @@
...
+        when(mockDocumentGeneratorService.generate(request)).thenThrow(RuntimeException.class);
+
+        ResponseEntity responseEntity = documentGeneratorController.generateDocument(request, mockBindingResult);
+        assertEquals(HttpStatus.INTERNAL_SERVER_ERROR, responseEntity.getStatusCode());
+    }
+
+    /**
+     * Set the data for the document generator request
+     *
+     * @return
```
BSaunders95: Can these return tags be populated please?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
BSaunders95: 👍

PR #22, opened 2018-09-10
Feature/document generator service implementation

Mathsh:
Creation of DocumentGeneratorServiceImpl to obtain the data from the DocumentInfoService and send to documentRenderService to generate a document

- Added DocumentGeneratorService interface and Impl
- Added DocumentRequest and Response models
- Updated RenderDocumentRequestHandlerTest as no longer returning status in RenderDocumentResponse
- Updated RenderDocumentRepsonse as no longer return status
- Added DocumentGeneratorServiceTests
- Updated document-generator-interface DocumentInfo model to contain all required parameters for return

Resolves SFA 664


document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorService.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.companieshouse.document.generator.core.service;
+
+import uk.gov.companieshouse.document.generator.core.service.models.DocumentRequest;
+import uk.gov.companieshouse.document.generator.core.service.models.DocumentResponse;
+
+public interface DocumentGeneratorService {
+
+    //TODO service interface created impl to be done as part of SFA 664
```
BSaunders95: This can be removed now

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorService.java
```diff
@@ -0,0 +1,16 @@
...
+    /**
+     * generate the document
+     *
+     * @param documentRequest
+     * @return
```
BSaunders95: Is there meant to be a return here?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
+
+import static uk.gov.companieshouse.document.generator.core.DocumentGeneratorApplication.APPLICATION_NAME_SPACE;
+
+public class DocumentGeneratorServiceImpl implements DocumentGeneratorService {
+
+    private DocumentInfoService documentInfoService;
+
+    private EnvironmentReader environmentReader;
```
BSaunders95: If a static EnvironmentReaderImpl is instantiated at the class level, we can have static Strings on the class to hold the environment variable details; that way, we don't end up reading env files on every method call. Something like:

```
public class Class {

   private static String DOCUMENT_RENDER_SERVICE_HOST = READER.getMandatoryString("DOCUMENT_RENDER_SERVICE_HOST");

}
```

The added benefit of this is that any `MissingEnvironmentVariableException`'s will be thrown on class instantiation, as opposed to when methods are called so we fail earlier.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+
+        this.documentInfoService = documentInfoService;
+        this.environmentReader = environmentReader;
+        this.requestHandler = requestHandler;
+    }
+
+    @Override
+    public DocumentResponse generate(DocumentRequest documentRequest) {
+
+        DocumentInfo documentInfo = null;
```
BSaunders95: Do these need to be instantiated as null explicitly?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        } catch (Exception e){
+            LOG.error(e);
+        }
+
+        if (documentInfo != null) {
+            renderResponse = renderSubmittedDocumentData(documentRequest, documentInfo);
+            response = setDocumentResponse(renderResponse, documentInfo);
+        }
+
+        if (response != null) {
```
BSaunders95: Don't need this check, just return `response`.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+
+        return null;
+    }
+
+    /**
+     * Send data to Render Service and generate document
+     *
+     * @param documentRequest
+     * @param documentInfo
+     * @return
```
BSaunders95: Should there be a return here?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+     * @param renderResponse
+     * @param documentInfo
+     * @return
+     */
+    private DocumentResponse setDocumentResponse(RenderDocumentResponse renderResponse, DocumentInfo documentInfo) {
+
+        DocumentResponse response;
+
+        response = new DocumentResponse();
+        response.setLocation(renderResponse != null ? renderResponse.getLocation() : null);
```
BSaunders95: Can we have a `renderResponse != null` check wrapping these setters? That way the values will just remain null rather than being explicitly set as null

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorServiceTest.java
```diff
@@ -0,0 +1,151 @@
...
+    @DisplayName("Test a successful generate completed")
+    public void testsSuccessfulGenerateCompleted() throws IOException {
+
+        when(mockDocumentInfoService.getDocumentInfo()).thenReturn(setSuccessfulDocumentInfo());
+
+        when(mockRequestHandler.sendDataToDocumentRenderService(any(String.class), any(RenderDocumentRequest.class))).thenReturn(setSuccessfulRenderResponse());
+
+        DocumentResponse response = documentGeneratorService.generate(setValidRequest());
+
+        assertNotNull(response);
```
BSaunders95: Should we perform an `assertEquals` here? I'm not fussy either way, just makes the tests a little more verbose. static strings could be held at the class level to hold the DocumentInfo and RenderResponse details

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorServiceTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertNotNull(response.getDescriptionValues());
+        assertNotNull(response.getLocation());
+        assertNotNull(response.getSize());
+    }
+
+    @Test
+    @DisplayName("Tests when an error thrown from documentInfoService")
+    public void testsWhenErrorThrownFromDocumentInfoService() {
+
+        when(mockDocumentInfoService.getDocumentInfo()).thenThrow(RuntimeException.class);
```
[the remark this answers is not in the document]
BSaunders95: We should be throwing checked exceptions if the exception is expected

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorService.java
Mathsh: Yes, it returns a Document Response

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
Mathsh: Yes

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
Mathsh: yes returns a RenderDocumentResponse so returns something

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
Mathsh: what's wrong with explicitly setting them to ensure they are null? It is important that they are. Don't want any instance of them not being.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorService.java
BSaunders95: Can the Javadoc be updated to reflect that please?

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorServiceTest.java
Mathsh: I just want to to check that the response is not null

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
Mathsh: problem with that is you can't mock a static variable. So that is why we are doing construction injection which is a Spring recommendation.

BSaunders95: Fair point, disregard this one then

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        requestData.setTemplateName(documentInfo.getTemplateName());
+        requestData.setLocation(documentInfo.getLocation());
+
+        try {
+            return requestHandler.sendDataToDocumentRenderService(url, requestData);
+        } catch (Exception e) {
+            LOG.error(e);
+        }
+
+        return null;
```
BSaunders95: Should we be returning null here? In the `catch` case I would assume that we'd cascade the exception to the controller

Mathsh: yes we should, need to still populate a response further on, but this will return null as it has failed to render document, so no location or size present.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfoResponse;
+import uk.gov.companieshouse.environment.EnvironmentReader;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
+
+import static uk.gov.companieshouse.document.generator.core.DocumentGeneratorApplication.APPLICATION_NAME_SPACE;
+
+public class DocumentGeneratorServiceImpl implements DocumentGeneratorService {
+
+    private DocumentInfoService documentInfoService;
```
BSaunders95: Should we be calling this from the controller? That way we only call `generate` if there's document info available

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+     * @return A populated RenderDocumentResponse model or Null
+     */
+    private RenderDocumentResponse renderSubmittedDocumentData(DocumentRequest documentRequest,
+                                                               DocumentInfoResponse documentInfoResponse) {
+
+        String host = environmentReader.getMandatoryString(DOCUMENT_RENDER_SERVICE_HOST_ENV_VAR);
+        String url = host + CONTEXT_PATH;
+
+        RenderDocumentRequest requestData = new RenderDocumentRequest();
+        requestData.setAssetId(documentInfoResponse.getAssetId());
```
BSaunders95: Could we use `BeanUtils.copyProperties` here?

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+        assertNotNull(response.getDescriptionIdentifier());
+        assertNotNull(response.getDescriptionValues());
+        assertNull(response.getLocation());
+        assertNull(response.getSize());
+    }
+
+    /**
+     * Set data for valid DocumentRequest
+     *
+     * @return
```
BSaunders95: Can return tags be populated please?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+        this.environmentReader = environmentReader;
+        this.requestHandler = requestHandler;
+    }
+
+    @Override
+    public DocumentResponse generate(DocumentRequest documentRequest) {
+
+        DocumentInfoResponse documentInfoResponse;
+        DocumentResponse response = null;
+        RenderDocumentResponse renderResponse;
```
jsimons1989: Can you instantiate these variables where they are needed please?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
jsimons1989: We agreed in the exception handling discussion that we would throw exceptions up to the controller. The exception in the catch block should be logged, wrapped in an exception relevant to this layer and thrown up to the controller to be converted into the correct HTTP status code.

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    private DocumentInfoResponse setSuccessfulDocumentInfo() {
+
+        DocumentInfoResponse documentInfoResponse = new DocumentInfoResponse();
+        documentInfoResponse.setLocation("location");
+        documentInfoResponse.setTemplateName("templateName");
+        documentInfoResponse.setData("data");
+        documentInfoResponse.setAssetId("assetId");
+        documentInfoResponse.setDescription("description");
+        documentInfoResponse.setDescriptionIdentifier("descriptionIdentifier");
```
jsimons1989: Can you make these into constants and verify with an assertEquals() instead of assertNull() in the tests please?

document-generator-interfaces/src/main/java/uk/gov/companieshouse/document/generator/interfaces/model/DocumentInfo.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Map;
+
+public class DocumentInfo {
+    
+    String data;
+    String templateName;
+    String assetId;
+    String location;
+    String Description;
+    String DescriptionIdentifier;
```
jsimons1989: These shouldn't be capitalised. Can you make the variables private to encapsulate them too please?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/service/impl/DocumentGeneratorServiceImpl.java
Mathsh: best to not have business logic in a controller. That's why it is in a service.

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/service/DocumentGeneratorServiceTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+        ResponseObject response = documentGeneratorService.generate(setValidRequest(), REQUEST_ID);
+
+        assertNotNull(response);
+        assertNotNull(response.getData());
+        assertEquals(response.getData().getDescriptionValues(), setDescriptionValue());
+        assertEquals(response.getData().getDescription(), DESCRIPTION);
+        assertEquals(response.getData().getDescriptionIdentifier(), DESCRIPTION_IDENTIFIER);
+        assertEquals(response.getData().getSize(), SIZE);
+        assertEquals(response.getData().getLinks(), LOCATION);
```
jsimons1989: These are the wrong way round. The method signature is: assertEquals(expected, actual);

document-generator-core/src/test/java/uk/gov/companieshouse/document/generator/core/utility/ApiResponseMapperTest.java
```diff
@@ -0,0 +1,109 @@
...
+        documentResponse.setDescriptionIdentifier(DESCRIPTION_IDENTIFIER);
+        documentResponse.setLinks(LOCATION);
+
+        ResponseObject responseObject = new ResponseObject(ResponseStatus.CREATED, documentResponse);
+
+        ResponseEntity responseEntity = apiResponseMapper.map(responseObject);
+
+        assertNotNull(responseEntity);
+        assertNotNull(responseEntity.getBody());
+        assertEquals(responseEntity.getStatusCode(), HttpStatus.CREATED);
```
jsimons1989: Assertions in this test class are the wrong way round. The method signature is: assertEquals(expected, actual);

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
```diff
@@ -10,53 +11,49 @@
...
 
-        DocumentResponse response;
+        String requestId = request.getHeader("X-Request-Id");
 
         if (result.hasErrors()) {
-            LOG.error("error in request body");
+            final Map<String, Object> debugMap = new HashMap<>();
+            debugMap.put("resource_uri", documentRequest.getResourceUrl());
+            debugMap.put("resource_id", documentRequest.getResourceId());
+            LOG.debug("error in request body", debugMap);
```
BSaunders95: Should this be an error log? If so, `LOG.errorRequest` can be used, taking `request` as a parameter, which adds the requestId to the json blob output in the logs which will help support in debugging.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/utility/ApiResponseMapper.java
```diff
@@ -0,0 +1,24 @@
...
+    public ResponseEntity map(ResponseObject responseObject) {
+
+        switch(responseObject.getStatus()) {
+            case CREATED:
+                return ResponseEntity.status(HttpStatus.CREATED).body(responseObject.getData());
+            case NO_DATA_RETRIEVED:
+                return ResponseEntity.status(HttpStatus.BAD_REQUEST).build();
+            case DOCUMENT_NOT_RENDERED:
+                return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(responseObject.getData());
+               default:
```
BSaunders95: Can this be bought inline with the other case statements please?

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/controller/DocumentGeneratorController.java
Mathsh: I checked previously with Lead, Log.debug was selected

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/utility/ApiResponseMapper.java
Mathsh: I'll do it on a future PR, as this has been approved and merged. before you commented. However, the default is only allowing me to place it in its current place in the IDE

PR #23, opened 2018-09-11
Adds skeleton implementation for Accounts resources

ChrisJBurns:
- adds AccountsHandler to handle accounts specific resources
- adds AccountsHandlerTest
- adds accounts specific scenario in AccountsDocumentInfoServiceImpl
- adds new scenario unit tests for AccountsDocumentInfoServiceImpl

NOTE: This PR will not cover the full handling of exception handling as this is still to be discussed as what throws what errors and where it gets caught. This PR simply lays out the skeleton of the functionality where the exception handling can be added on top.

Implements: SFA-567


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -34,27 +38,32 @@ public DocumentInfo getDocumentInfo() {
...
                 .map(Resource::getLinks)
                 .map(links -> links.get("resource"))
-                // when abridged has been migrated to use the company-accounts api, the code for the
-                // company accounts should work for abridged, resulting in this abridged specific code
-                // qualifying for removal
-                .filter(this::isAbridged)
-                .map(accountsLinks -> new DocumentInfo())
-                .orElse(null);
+                .orElseGet(() -> {
+                    LOG.info("Unable to find resource in transaction: " + resourceId);
```
jsimons1989: Can you log the transaction ID here too please.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -34,27 +38,32 @@ public DocumentInfo getDocumentInfo() {
...
+                    return "";
+                });
+
+
+        // when the Accounts migration has been completed to Company Accounts, this code can be removed
+        if (isAccounts(resourceLink)) {
+            return accountsHandler.getAccountsData(resourceLink);
+         }
 
+        return null;
```
jsimons1989: I know this is a skeleton and you are trying to keep the PR small but errors should be handled appropriately. You are returning a mixture of null and "" returns in this class.

At this point an OperationUnsupported exception could be thrown to clearly show that other account types are not yet supported. This would make it clear where the code should be extended to support other account types.

The other error returns in this class are all caused by the user supplying a URL to missing/incomplete data, I believe from a previous conversation with the architects that this will cause a 400 error. The return type/exception should be consistent and I believe should be in this PR.

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImplTest.java
```diff
@@ -27,40 +28,47 @@
...
     @Test
     @DisplayName("Tests the unsuccessful retrieval of an accounts document data due to no accounts resource in transaction")
     void testUnsuccessfulGetDocumentInfoNoAccountsResourceInTransaction() {
-        Map<String, Resource> resources = new HashMap<>();
-        resources.put("error", createResource());
-
-        Transaction transaction = new Transaction();
-        transaction.setResources(resources);
+        Transaction transaction = createTransaction();
+        transaction.getResources().remove("");
```
jsimons1989: Can you use a test ID instead of an empty string please? It'll be easier to see the use of the resource ID in the tests if it's a constant instead of duplicating the empty string.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: Done

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
ChrisJBurns: As mentioned in the PR description the exception handling hasn't been design or discussed yet so I will leave it until that has happened

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -26,25 +27,25 @@
...
+    public DocumentInfoResponse getDocumentInfo(DocumentInfoRequest documentInfoRequest) {
         LOG.info("Started getting document");
 
-        String resource = "";
-        String resourceId = "";
+        String resourceId = documentInfoRequest.getResourceId();
+        String resourceUri = documentInfoRequest.getResourceUri();
 
-        Transaction transaction = transactionService.getTransaction(resource);
+        Transaction transaction = transactionService.getTransaction(resourceId);
```
jsimons1989: This should be the resourceURI. The resource URI is the self link in the transaction.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -26,25 +27,25 @@
...
         if (transaction == null) {
             LOG.error("transaction not found");
             return null;
         }
 
         String resourceLink =  Optional.of(transaction)
                 .map(Transaction::getResources)
-                .map(resources -> resources.get(resourceId))
+                .map(resources -> resources.get(resourceUri))
```
jsimons1989: This should be the resourceId. It's the ID of the resource within the transaction.

PR #25, opened 2018-09-12
Adds document info request in unit tests

ChrisJBurns:
- fixes unit tests
- adds request param into the implementation
- additional implementation will be added in #23

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
@@ -56,6 +56,6 @@ public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
      * @return true if abridged, false if not
      */
     private boolean isAbridged(String accountLink) {
-        return accountLink.matches("/transactions\\/[0-9\\-]+\\/accounts\\/.*");
+        return accountLink.matches("/transactions\\/[0-9-]+/accounts\\/.*");
     }
 }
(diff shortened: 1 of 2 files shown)


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -55,6 +56,6 @@ public DocumentInfo getDocumentInfo() {
      * @return true if abridged, false if not
      */
     private boolean isAbridged(String accountLink) {
-        return accountLink.matches("/transactions/\\d+(\\-\\d+)+/accounts/\\w+?=");
+        return accountLink.matches("/transactions\\/[0-9\\-]+\\/accounts\\/.*");
```
BSaunders95: You shouldn't need the escape of the `-` inside the block. Intellij tends to complain about this anyway.

ChrisJBurns: fixed

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImplTest.java
```diff
@@ -30,6 +32,9 @@
     @Mock
     private TransactionService transactionService;
 
+    private static final String RESOURCE_ID = "091174-913515-326060";
```
BSaunders95: Could these be made more generic? Still follow the same regex pattern, but I feel like unit tests should be as 'fake' as possible.

ChrisJBurns: Good point however this PR simply addresses the broken tests, any improvement will can be made further on

BSaunders95: I'll approve for now, but I'd advise this be done at some point. Even though it has no bearing on the tests, it all feels a little _too_ real.

Issue #14, opened 2018-08-30
Check for mandatory environment variables on startup


ChrisJBurns:
Currently we do not check for mandatory environment variables on start up that are fundamental to the document generator working. We should insert a check that identifies whether or not the `API_URL` and `CHS_API_KEY` are defined as environment variables.

Additionally when the SDK is integrated the way the environment variables is loaded will need to be revisited. Need to confirm what the SDK is doing and what our service is expected to do.

ChrisJBurns:
Hi @jsimons1989 long time no speak :P do you mind if I close this? I'm just going through my GitHub account as part of normal housekeeping and am seeing some CH issues that I opened a while back that are still open and showing up on my profile. There are others too but depending on your response to this issue depends what I do to the others :)

PR #27, opened 2018-09-12
Adds accounts service layer

ChrisJBurns:
- adds accounts service layer that interacts with the currently implemented internal `private-sdk`. When the `private-sdk` has been completed, the accounts service layer replaces all references to the internal temporary `private-sdk` solution with the real one.

All refined exception handling and logging will be discussed and designed later on and can be easily added. That is not the scope of this PR. This PR simply deals with the skeleton basic accounts service layer and some basic exception handling and logging.

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
@@ -1,20 +1,30 @@
 package uk.gov.companieshouse.document.generator.accounts.data.accounts;
 
+import static uk.gov.companieshouse.document.generator.accounts.AccountsDocumentInfoServiceImpl.MODULE_NAME_SPACE;
+
+import java.util.HashMap;
+import java.util.Map;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.stereotype.Component;
 import org.springframework.web.client.RestTemplate;
+import uk.gov.companieshouse.api.error.ApiErrorResponseException;
 import uk.gov.companieshouse.api.model.accounts.Accounts;
 import uk.gov.companieshouse.environment.EnvironmentReader;
 import uk.gov.companieshouse.environment.impl.EnvironmentReaderImpl;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
 
 /**
  * Temporary solution until private-sdk has been completed (SFA-518, SFA-670). When completed, this
  * file will get removed alongside the data package and all references to this file will be replaced
  * with calls to the private-sdk.
  */
+@Component
 public class AccountsManager {
 
     /** represents the Authorization header name in the request */
@@ -22,10 +32,12 @@ public class AccountsManager {
 
     private static final EnvironmentReader READER = new EnvironmentReaderImpl();
 
-    private static String API_URL = READER.getMandatoryString("API_URL");
-    private static String CHS_API_KEY = READER.getMandatoryString("CHS_API_KEY");
+    private final String API_URL = READER.getMandatoryString("API_URL");
+    private final String CHS_API_KEY = READER.getMandatoryString("CHS_API_KEY");
 
     private static final RestTemplate restTemplate = createRestTemplate();
+
+    private static final Logger LOG = LoggerFactory.getLogger(MODULE_NAME_SPACE);
 
     /**
      * Get accounts resource if exists
@@ -33,12 +45,26 @@ public class AccountsManager {
      * @param link - self link for the accounts object
      * @return accounts object along with the status or not found status.
      */
-    public static ResponseEntity<Accounts> getAccounts(String link) {
+    public Accounts getAccounts(String link) throws Exception {
         HttpHeaders requestHeaders = new HttpHeaders();
         requestHeaders.set(AUTHORIZATION_HEADER, getApiKey());
 
         HttpEntity requestEntity = new HttpEntity(requestHeaders);
-        return restTemplate.exchange(getRootUri() + link, HttpMethod.GET, requestEntity, Accounts.class);
+
+        ResponseEntity<Accounts> accountsResponseEntity = restTemplate.exchange(getRootUri() + link, HttpMethod.GET, requestEntity, Accounts.class);
+
+        if (accountsResponseEntity.getStatusCode() != HttpStatus.OK) {
+            Map<String, Object> logMap = new HashMap<>();
+            logMap.put("resource", link);
+            logMap.put("data.status", accountsResponseEntity.getStatusCode());
+            LOG.error("Failed to retrieve data from API", logMap);
+
+            // we are throwing a generic exception to mimic the private sdk throwing an exception.
+            // We're not to create a custom exception as it will have to be removed when the private sdk
+            // gets implemented - additionally the generic exception is sufficient
+            throw new Exception("Failed to retrieve data from API");
+        }
+        return accountsResponseEntity.getBody();
     }
 
     /**
@@ -51,11 +77,11 @@ public class AccountsManager {
         return new RestTemplate(requestFactory);
     }
 
-    private static String getRootUri() {
+    private String getRootUri() {
         return API_URL;
     }
 
-    private static String getApiKey() {
+    private String getApiKey() {
         return CHS_API_KEY;
     }
 }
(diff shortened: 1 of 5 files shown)


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(MODULE_NAME_SPACE);
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Accounts getAccounts(String resource) {
+        LOG.info("Getting accounts data");
+
+        ResponseEntity<Accounts> accounts = AccountsManager.getAccounts(resource);
```
jsimons1989: The private-sdk won't be returning a ResponseEntity. See get() in [AccountResourceHandler](https://github.com/companieshouse/api-sdk-java/blob/master/src/main/java/uk/gov/companieshouse/api/handler/transaction/account/AccountResourceHandler.java). The AccountsManager should return the Accounts object or throw a custom exception to represent the ApiErrorResponseException thrown by the SDK.

If you remove the static from the methods in the AccountsManager it can then be mocked for unit testing which means the unit tests can be used with minimal modification (only need to replace the mock) when the private SDK is integrated.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+public class AccountsServiceImpl implements AccountsService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MODULE_NAME_SPACE);
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Accounts getAccounts(String resource) {
+        LOG.info("Getting accounts data");
```
jsimons1989: This log should contain the resource too

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+        ResponseEntity<Accounts> accounts = AccountsManager.getAccounts(resource);
+
+        if (accounts.getStatusCode() != HttpStatus.OK) {
+            LOG.error("Failed to retrieve data from API: " + resource +
+            "Status Code: " + accounts.getStatusCode().value());
```
jsimons1989: Can you use the same keys as other applications so this gets displayed in a consistent way (data.status) in our logging UI please? I am happy to demo what I mean.

ChrisJBurns: This shall not be needed then as the `ResponseEntity` will not be returned.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/service/impl/AccountsServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+        Accounts accounts;
+        try {
+            accounts = accountsManager.getAccounts(resource);
+        } catch (Exception e) {
+            LOG.error(e);
+            throw new ServiceException(e.getMessage(), e.getCause());
+        }
+
+        LOG.trace("Accounts data retrieved successfully: " + resource);
+        return accounts;
```
BSaunders95: The return could be performed inside the `try` clause to save assigning a temporary variable `accounts`

ChrisJBurns: It was more for the logging on line 37, but I'll remove it for now and do the one line return as the logging is still to be refined.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -1,44 +1,70 @@
...
+            logMap.put("resource", link);
+            logMap.put("data.status", accountsResponseEntity.getStatusCode());
+            LOG.error("Failed to retrieve data from API", logMap);
+
+            // we are throwing a generic exception to mimic the private sdk throwing an exception.
+            // We're not to create a custom exception as it will have to be removed when the private sdk
+            // gets implemented - additionally the generic exception is sufficient
+            throw new Exception("Failed to retrieve data from API");
+        }
+        return accountsResponseEntity.getBody();
```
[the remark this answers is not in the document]
ChrisJBurns: nothing is return in that case, it shall throw an exception

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/exception/ServiceException.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.companieshouse.document.generator.accounts.exception;
+
+public class ServiceException extends Exception {
+
+    public ServiceException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
```
Mathsh: white space

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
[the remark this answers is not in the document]
jterrych: ah yes, its right there and i didnt see it. Even so, I still think you have the test to enforce the expected Exception, and move the comment to the doc. The comment will be forever otherwise, smelling up the code with a generic Exception that will never be refactored when it should.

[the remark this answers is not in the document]
ChrisJBurns: It is worth noting that this file is just an internal solution until the `private-sdk-java` gets created. When that has been done, this file will get deleted along with a few others and they will be replaced with just method calls to the `private-sdk`.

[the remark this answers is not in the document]
jterrych: Chips is full of internal temp solutions too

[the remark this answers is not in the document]
ChrisJBurns: What would you suggest? Put the comment in the JD or remove it completely?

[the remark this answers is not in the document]
jterrych: JD should reference whats being thrown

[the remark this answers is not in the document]
ChrisJBurns: Done :+1:

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -1,44 +1,70 @@
...
 
     /** represents the Authorization header name in the request */
     private static final String AUTHORIZATION_HEADER = "Authorization";
 
     private static final EnvironmentReader READER = new EnvironmentReaderImpl();
 
-    private static String API_URL = READER.getMandatoryString("API_URL");
-    private static String CHS_API_KEY = READER.getMandatoryString("CHS_API_KEY");
+    private final String API_URL = READER.getMandatoryString("API_URL");
+    private final String CHS_API_KEY = READER.getMandatoryString("CHS_API_KEY");
```
jsimons1989: If this is final it can only be instantiated once per instance created. Why don't you make it static so it only needs to get instantiated once? We should be loading the config when the service starts to fail the start-up if config is missing.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -1,44 +1,70 @@
...
 
         HttpEntity requestEntity = new HttpEntity(requestHeaders);
-        return restTemplate.exchange(getRootUri() + link, HttpMethod.GET, requestEntity, Accounts.class);
+
+        ResponseEntity<Accounts> accountsResponseEntity = restTemplate.exchange(getRootUri() + link, HttpMethod.GET, requestEntity, Accounts.class);
+
+        if (accountsResponseEntity.getStatusCode() != HttpStatus.OK) {
+            Map<String, Object> logMap = new HashMap<>();
+            logMap.put("resource", link);
+            logMap.put("data.status", accountsResponseEntity.getStatusCode());
```
jsimons1989: You only need to use the key "status" here. "data" will automatically get prefixed.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
jterrych: Also, all caps snake case is the convention for static final variables. It should be camel case for final instance variables. While being a minor nitpick, it does add smell to sonar.

ChrisJBurns: I tried to make it static but the test fail as `Mockito` falls over. Also @jterrych  good spot, my IDE highlighted it but must have forgot

jsimons1989:
Issue has been raised for the env variables: https://github.com/companieshouse/document-generator/issues/14

PR #28, opened 2018-09-17
Document Generator Structure Tidy

Mathsh:
A number of small alterations to correct issues with start.sh and MakeFile. Also corrected LOG.Debug to LOG.DebugRequest

Added Beans to ApplicationConfiguration for required Impls
Added @Service to ServiceImpl
Updated MakeFile to produce correct JAR and ZIP

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ all: build
 .PHONY: clean
 clean:
 	mvn clean
-	rm -f ./$(artifact_core-name)/$(artifact_core-name).jar
 	rm -f ./$(artifact_name).jar
 	rm -f ./$(artifact_name)-*.zip
 	rm -rf ./build-*
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,8 @@ all: build
 .PHONY: clean
 clean:
 	mvn clean
-	rm -f ./$(artifact_core-name).jar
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -11,15 +11,16 @@ all: build
 .PHONY: clean
 clean:
 	mvn clean
-	rm -f ./$(artifact_core-name).jar
+	rm -f ./$(artifact_core-name)/$(artifact_core-name).jar
```
jsimons1989: Will this ever exist? The core artifact is copied to ./$(artifact_name).jar only.

document-generator-core/src/main/java/uk/gov/companieshouse/document/generator/core/DocumentGeneratorApplicationConfiguration.java
```diff
@@ -12,4 +18,19 @@
...
+    }
+
+    @Bean
+    TransactionService transactionService() {
+        return new TransactionServiceImpl();
+    }
+
+    @Bean
+    AccountsHandler accountsHandler() {
+        return new AccountsHandlerImpl();
```
jsimons1989: Probably worth looking into component scanning in Spring in the future to prevent needing the beans defined and might make the autowiring easier to mock in the unit tests.

PR #32, opened 2018-09-18
Adds AccountsHandler implementation for base accounts resource

ChrisJBurns:
- adds `ServiceException` constructor accepting just a string error message
- adds `HandlerException` for the handler implementations
- adds `AccountType` enum 
- adds `getAccounts` resource implementation in `AccountsHandler`
- adds and improves unit test coverage around the new scenarios and implementation

NOTE: This PR does not cover the fine tuned Exception handling or logging. It simply puts in place a skeleton in which we can build on afterwards.

Resolves: SFA-567

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
@@ -5,6 +5,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import uk.gov.companieshouse.api.model.transaction.Resource;
 import uk.gov.companieshouse.api.model.transaction.Transaction;
+import uk.gov.companieshouse.document.generator.accounts.exception.HandlerException;
 import uk.gov.companieshouse.document.generator.accounts.handler.accounts.AccountsHandler;
 import uk.gov.companieshouse.document.generator.accounts.service.TransactionService;
 import uk.gov.companieshouse.document.generator.interfaces.DocumentInfoService;
@@ -52,8 +53,12 @@ public class AccountsDocumentInfoServiceImpl implements DocumentInfoService {
 
         // when the Accounts migration has been completed to Company Accounts, this code can be removed
         if (isAccounts(resourceLink)) {
-            return accountsHandler.getAccountsData(resourceLink);
-         }
+            try {
+                return accountsHandler.getAccountsData(resourceLink);
+            } catch (HandlerException e) {
+                LOG.error(e);
+            }
+        }
 
         return null;
     }
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
@@ -1,13 +1,55 @@
 package uk.gov.companieshouse.document.generator.accounts.handler.accounts;
 
+import static uk.gov.companieshouse.document.generator.accounts.AccountsDocumentInfoServiceImpl.MODULE_NAME_SPACE;
+
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.model.accounts.Accounts;
+import uk.gov.companieshouse.document.generator.accounts.AccountType;
+import uk.gov.companieshouse.document.generator.accounts.exception.HandlerException;
+import uk.gov.companieshouse.document.generator.accounts.exception.ServiceException;
+import uk.gov.companieshouse.document.generator.accounts.service.AccountsService;
 import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfoResponse;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
 
 @Component
 public class AccountsHandlerImpl implements AccountsHandler  {
 
+    private static final Logger LOG = LoggerFactory.getLogger(MODULE_NAME_SPACE);
+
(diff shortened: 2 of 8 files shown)


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
```diff
@@ -52,8 +53,12 @@ public DocumentInfoResponse getDocumentInfo(DocumentInfoRequest documentInfoRequ
 
         // when the Accounts migration has been completed to Company Accounts, this code can be removed
         if (isAccounts(resourceLink)) {
-            return accountsHandler.getAccountsData(resourceLink);
-         }
+            try {
+                return accountsHandler.getAccountsData(resourceLink);
+            } catch (HandlerException e) {
+                LOG.error(e);
```
BSaunders95: Should this log be more detailed to aid support?

ChrisJBurns: Check the description of the PR:
>NOTE: This PR does not cover the fine tuned Exception handling or logging. It simply puts in place a skeleton in which we can build on afterwards.

;)

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -1,13 +1,56 @@
...
+
     @Override
-    public DocumentInfoResponse getAccountsData(String resourceLink) {
+    public DocumentInfoResponse getAccountsData(String resourceLink) throws HandlerException {
+        Accounts accounts;
+
+        try {
+            accounts = accountsService.getAccounts(resourceLink);
+        } catch (ServiceException e) {
+            LOG.error(e);
```
BSaunders95: No need to log here; this exception is already logged at the controller

ChrisJBurns: removed

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImpl.java
BSaunders95: My mistake!

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -1,13 +1,55 @@
...
         return new DocumentInfoResponse();
     }
+
+
+    /**
+     * Get the account type from the account link within links
+     *
+     * @return accountsData - accounts data
+     */
+    private AccountType getAccountType(Accounts accountsData) throws HandlerException {
```
BSaunders95: Should this method be part of the `AccountType` enum?

ChrisJBurns: There is already a method that gets the account type from the enum, this one specifically gets the account type from the accounts data set. Both do different things

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -1,13 +1,55 @@
...
+        return accountsData.getLinks().keySet()
```
jterrych: you shouldnt iterate over the keyset to check if the key exists in a map. Just use Optional instead, if the key get returns a null then it will still throw the exception you expect.
```
return Optional.ofNullable(accountsData.getLinks())
                          .map(links -> links.get(LinkType.SELF.getLink()))
                          .map(AccountType::getAccountType)
                          .orElseThrow(...);
```

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
ChrisJBurns: To be a bit more specific.
- The `getAccountType` in the enum simply gets the account type from the enum given the key
- The `getAccountType` in the Handler gets the account type from the accounts data set by iterating over the enum `getAccountType` for each element until it matches

BSaunders95: That's fine, but it feels as though it would be the concern of the enum to return an AccountType for any possible input. The 2 methods could both exist in the enum, and then be javadoc'd appropriately.

ChrisJBurns: I'm not sure, enums normally only concern themselves with the logic needed internally, they shouldn't really be connecting other working classes of the application to it. Additionally as we are currently looking through the accounts data, would we have to create a new method for a different type of accounts? This then gets larger and larger etc

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
ChrisJBurns: Optional doesn't work as we are searching for LinkTypes that aren't SELF link.

jterrych: yep my bad.

PR #34, opened 2018-09-18
Adds Abridged accounts implementation in handler

ChrisJBurns:
- adds Abridged accounts implementation in AccountsHandler
- updates the references of the new abridged method in the controller class
- adds and improves the unit test coverage of the new scenarios
- improves javaDocs where needed

Resolves: SFA-567

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
@@ -5,6 +5,7 @@ import static uk.gov.companieshouse.document.generator.accounts.AccountsDocument
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import uk.gov.companieshouse.api.model.accounts.Accounts;
+import uk.gov.companieshouse.api.model.accounts.abridged.AbridgedAccountsApi;
 import uk.gov.companieshouse.document.generator.accounts.AccountType;
 import uk.gov.companieshouse.document.generator.accounts.LinkType;
 import uk.gov.companieshouse.document.generator.accounts.exception.HandlerException;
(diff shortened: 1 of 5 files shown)


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -32,17 +36,27 @@ public DocumentInfoResponse getAccountsData(String resourceLink) throws HandlerE
...
-        // TODO: The accountsType variable shall be used in the implementation of abridged logic
         AccountType accountsType = getAccountType(accounts);
 
+        String abridgedAccountLink = getAccountLink(accounts, accountsType);
+        try {
+            // TODO: abridgedAccountData will be used as part of the implementation of converting
+            // TODO: the data object to a json string
+            AbridgedAccountsApi abridgedAccountData = accountsService.getAbridgedAccounts(abridgedAccountLink);
+        } catch (ServiceException e) {
+            throw new HandlerException(e.getMessage(), e.getCause());
```
jsimons1989: Can you log a message here with the available details of the account that was affected by this error (account ID, account type, etc) please?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -32,17 +36,27 @@ public DocumentInfoResponse getAccountsData(String resourceLink) throws HandlerE
             throw new HandlerException(e.getMessage(), e.getCause());
         }
 
-        // TODO: The accountsType variable shall be used in the implementation of abridged logic
         AccountType accountsType = getAccountType(accounts);
 
+        String abridgedAccountLink = getAccountLink(accounts, accountsType);
```
jsimons1989: The account type is being retrieved from the links to then retrieve the link using the account type. Is the account type going to be used for anything else or can this be simplified to just get the non-self link?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
ChrisJBurns: Refined logging will be done in a separate PR as explain in the PR description as well as the refined exception handling. I have submitted basic logging so this should be fine, but remember that the logging isn't going to be complete in these PRs until there is a PR that addressing logging entirely.

jsimons1989: Logging is a non-functional requirement that should exist in each PR. I agree that we should review/improve it over time as we use the logs and it won't be perfect the first time but it needs to be included.

PR #36, opened 2018-09-19
Implements functionality around response data

ChrisJBurns:
- adds response data by getting information from abridged data and account type
- adds and improves coverage of unit tests for new code and scenarios
- corrects JavaDoc's on `HandlerException` and `ServiceException`

Resolves: SFA-567

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/exception/HandlerException.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/exception/HandlerException.java
@@ -1,11 +1,29 @@
 package uk.gov.companieshouse.document.generator.accounts.exception;
 
+/**
+ * The class {@code HandlerException} is a form of {@code Exception}
+ * that should be used at the handler layer to abstract lower level
+ * level exceptions from being propagated up the call stack.
+ */
 public class HandlerException extends Exception {
 
+    /**
+     * Constructs a new {@code HandlerException} with a custom message.
+     *
+     * @param message a custom message
+     */
     public HandlerException(String message) {
         super(message);
     }
 
+
+    /**
+     * Constructs a new {@code HandlerException} with a custom message and the specified
+     * cause.
+     *
+     * @param message a custom message
+     * @param cause the cause
+     */
     public HandlerException(String message, Throwable cause) {
         super(message, cause);
     }
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/exception/HandlerException.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/exception/HandlerException.java
@@ -1,14 +1,14 @@
 package uk.gov.companieshouse.document.generator.accounts.exception;
 
 /**
- * The class {@code HandlerException} is a form of {@code Exception}
- * that should be used at the handler layer to abstract lower level
- * level exceptions from being propagated up the call stack.
+ * HandlerException is a wrapper exception that hides
+ * lower level exceptions from the caller and prevents them
+ * from being propagated up the call stack.
  */
 public class HandlerException extends Exception {
 
     /**
-     * Constructs a new {@code HandlerException} with a custom message.
+     * Constructs a new HandlerException with a custom message.
      *
      * @param message a custom message
      */
@@ -18,7 +18,7 @@ public class HandlerException extends Exception {
 
 
     /**
-     * Constructs a new {@code HandlerException} with a custom message and the specified
+     * Constructs a new HandlerException with a custom message and the specified
      * cause.
      *
      * @param message a custom message
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandler.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandler.java
@@ -1,5 +1,6 @@
 package uk.gov.companieshouse.document.generator.accounts.handler.accounts;
 
+import uk.gov.companieshouse.api.model.transaction.Transaction;
 import uk.gov.companieshouse.document.generator.accounts.exception.HandlerException;
 import uk.gov.companieshouse.document.generator.interfaces.model.DocumentInfoResponse;
 
@@ -13,8 +14,9 @@ public interface AccountsHandler {
 
     /**
      * Get an Abridged accounts resource from the given resource link
+     * @param transaction the transaction data
      * @param resourceLink the resource link of the abridged accounts
      * @return a populated {@link DocumentInfoResponse} object
      */
-    DocumentInfoResponse getAbridgedAccountsData(String resourceLink) throws HandlerException;
+    DocumentInfoResponse getAbridgedAccountsData(Transaction transaction, String resourceLink) throws HandlerException;
 }
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandler.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandler.java
@@ -17,6 +17,7 @@ public interface AccountsHandler {
      * @param transaction the transaction data
      * @param resourceLink the resource link of the abridged accounts
      * @return a populated {@link DocumentInfoResponse} object
+     * @throws HandlerException throws a custom handler exception
      */
     DocumentInfoResponse getAbridgedAccountsData(Transaction transaction, String resourceLink) throws HandlerException;
 }
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
@@ -127,5 +127,4 @@ public class AccountsHandlerImpl implements AccountsHandler  {
     private String getBucketName() {
         return bucketName;
     }
-    
 }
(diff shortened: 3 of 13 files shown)


ChrisJBurns:
Build fixed

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -83,5 +88,43 @@ private AccountType getAccountType(Accounts accountsData) throws HandlerExceptio
...
+        DocumentInfoResponse documentInfoResponse = new DocumentInfoResponse();
+        documentInfoResponse.setData(createDocumentInfoResponseData(transaction, abridgedAccountData));
+        documentInfoResponse.setAssetId(accountType.getAssetId());
+        documentInfoResponse.setTemplateName(accountType.getTemplateName());
+        documentInfoResponse.setLocation(createLocationString(accountType));
+        return documentInfoResponse;
+    }
+
+    private String createLocationString(AccountType accountType) {
+        return String.format("%s/%s/%s", getBucketName(), accountType.getAssetId(), accountType.getUniqueFileName());
```
jsimons1989: When is the prefix "s3://" being appended to the location?

ChrisJBurns:
@jsimons1989  I completely agree and have been taking note as I go along to what component does what etc. The only reason I haven't made this public is because things have changed along the way so I saw it as counterproductive updating something to something that may not be right - and with someone watching that space it may just confuse them with the change's that are going on. But yes, I will be putting all of that into a confluence page the second this PR and #38 is complete.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
ChrisJBurns: So this uses the `DOC_GEN_ACC_BUCKET_NAME` variable which will end up being: `DOC_GEN_ACC_BUCKET_NAME` = $STORAGE_TYPE://$DOCUMENT_RENDER_SERVICE_HOST`

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -83,5 +88,43 @@ private AccountType getAccountType(Accounts accountsData) throws HandlerExceptio
...
+    }
+
+    private String getBucketName() {
+        return bucketName;
```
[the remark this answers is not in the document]
ChrisJBurns: I completely agree, I did make the variable mandatory however, unfortunately the tests fail to run because the `DOC_GEN_ACC_BUCKET_NAME` doesn't exist. I would welcome suggestions to get around that. Normally we have put the variables in the maven config, however because of the modules, I'm wondering if we should be putting `document-generator-accounts` environment variables in the `document-generator`

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
jsimons1989: Have you raised a story for updating the configs when we make the switch to this new repo?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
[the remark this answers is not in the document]
jsimons1989: document-generator shouldn't contain the pluggable-code's environment variables. We'll be mixing generic and specific which could cause difficulty in the future updating or removing them. Without having spent time thinking about it, my quick suggestion would be an initialise method in the DocumentInfo interface to allow each implementation to initialise the configs it needs. This will keep the separation and allow failing early (service start-up) if the configs are not available. We would need to discuss with @Mathsh.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
ChrisJBurns: What would you recommend calling it? The story has been created as SFA-725. Will obviously change when we agree on a name for it.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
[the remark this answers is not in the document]
ChrisJBurns: As discussed offline, we will set the location in `doc-gen-api` and we will only return the path which does not need the `bucketName`

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
ChrisJBurns: As discussed offline, we will be using the `bucketName` in doc-gen-api and we will only return the path which does not need the `bucketName` so this PR will not need to worry about it.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandler.java
```diff
@@ -13,8 +14,10 @@
 
     /**
      * Get an Abridged accounts resource from the given resource link
+     * @param transaction the transaction data
      * @param resourceLink the resource link of the abridged accounts
      * @return a populated {@link DocumentInfoResponse} object
+     * @throws HandlerException throws a custom handler exception
```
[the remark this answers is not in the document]
ChrisJBurns: Yeah that was auto generated by my IDE - however I will JavaDoc the HandlerException, good spot :+1:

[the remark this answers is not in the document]
ch-tstrickland: Also there for a few of the ServiceExceptions - why thrown, what should I do?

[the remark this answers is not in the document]
ch-tstrickland: On the other hand I really liked that HandlerException does not have any no-message constructors - very nice, rarely done. Exceptions without messages are just rude.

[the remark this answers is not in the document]
ChrisJBurns: I have added JavaDoc for the `HandlerException` as I know what to do when it is thrown please let me know if the JavaDoc I have added is suffice enough as your eyes are fresh to this code.

Also, I agree, Exceptions without messages are hard to debug as you can't search for an error string 👎

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/exception/HandlerException.java
```diff
@@ -1,11 +1,29 @@
 package uk.gov.companieshouse.document.generator.accounts.exception;
 
+/**
```
ch-tstrickland: Several points:
'The class...' is superfluous. This is a class doc. Same applies to 'is a form of exception' as that is in the name.
How about:L
Wrapper exception that hides lower level exceptions from the caller.

That said, it is still not clear either from this Javadoc or from the method Javadocs what you should do/are allowed to do if you see this exception. It needs to be clear in one of these places, and I suspect it is the method docs.

ch-tstrickland: Same points apply to ServiceException too.
Also, no need for the link as you are on the page that it links to. Furthermore, in any Javadocs where type is there in the return or a parameter, a link will be generated and you do not need to make another one right next to it, as it just clutters the text.
If you look at the method Javadcos here, you'll see what I mean:
https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html

ChrisJBurns: I have altered the text and have pushed it but left the one last just to show the reason for why we have the custom exception.

ChrisJBurns: Additionally, I have only now just figured out the magic of ctrl + J in intelliJ where you can view the JD when it gets rendered. So am slowly going through correcting them :P

ch-tstrickland: Following a chat elsewhere, I take back some of my comments about instructing the caller on what to do. That point can sometimes apply, but not here.
Specifically, I was applying a previous lesson from a stateful service that threw exceptions in situations that sometimes implied that the method failed, but you still had a session, or sometimes the session had gone, or sometimes the service was generally somewhere up a creek... and the API could have made all this clear to the client but did not. None of that applies here: the method simply failed!

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandler.java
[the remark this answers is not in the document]
ChrisJBurns: It has been corrected I believe by Shaun within the last few minutes under #39  :+1:

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/exception/HandlerException.java
ChrisJBurns: Ahh I see, nah I don't think this applies here, it may be hard to see currently what to do, but it will make more sense when the refined exception handling has been done as there will be exceptions for certain things. Additionally, the design will be updated soon to take into account what the accounts implementation will do. Which in simple terms would be the following:

ServiceException throw -> caught in Handler layer, logs service exception details, throws new HandlerException -> caught in controller layer, logs the handlerexception, throws controller exception etc.

PR #45, opened 2018-10-10
Document Generator Complete logic around returning correct data to doc-gen-api

Mathsh:
Completed logic around returning correct data to doc-gen-api.
Added git submodule api-enumerations to document-generator-api in order to obtain the document description values.

- Updated .gitmodules to include api-enumerations within document-generator-api
- Updated MakeFile to build the submodule
- Created new service to retrieve the description from the api-enumerations
- Updated methods in document-generator-accounts to accept generic documentData
- Added getDescription method to document-generator-api to call new service.
- added methods to format date and obtain period end on date. 

Resolves SFA 724


document-generator-api/src/main/java/uk/gov/companieshouse/document/generator/api/document/description/impl/RetrieveApiEnumerationDescriptionImpl.java
```diff
@@ -0,0 +1,89 @@
...
+                                               Map<String, String> parameters) throws IOException {
+
+        Yaml yaml = new Yaml();
+        File descriptionsFile = new File(fileName);
+
+        String description = "";
+
+        InputStream inputStream;
+        try {
+            inputStream = new FileInputStream(descriptionsFile);
```
sroberts2: I would change this to a Try with resource so that you don't have to explicitly handle closing the Stream if an error occurs. So something like `try (InputStream inputStream = new FileInputStream(descriptionsFile)) {`

document-generator-api/pom.xml
```diff
@@ -81,7 +85,7 @@
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.8.0</version>
+            <version>${gson.version}</version>
```
sroberts2: This dependency is now a duplicate of the one above change it back or remove one.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -86,38 +105,75 @@ private String getAccountLink(Accounts accounts, AccountType accountsType) {
...
         documentInfoResponse.setTemplateName(accountType.getTemplateName());
         documentInfoResponse.setPath(createPathString(accountType));
+
+        Map<String, String> descriptionValues = new HashMap<>();
+        descriptionValues.put("period_end_on", RESPONSE_DISPLAY_DATE_FORMAT.format(ISO_DATE_FORMAT.
+                parse(getCurrentPeriodEndOn(accountData))));
+
+        documentInfoResponse.setDescriptionValues(descriptionValues);
+        documentInfoResponse.setDescriptionIdentifier(accountType.getEnumerationKey());
+        documentInfoResponse.setContentType(mimeType);
```
jsimons1989: I thought the MIME type was going to be set to the MIME type header in the response from the data API request (e.g. abridged accounts API)? This could result in a different content type to the data type.

jsimons1989: Is the Content-Type header allowed to be set to something other than the supplied MIME type (the document-specific code could set the Content-Type to anything and ignore the MIME type parameter)? Looking at the [document-render-service](https://github.com/companieshouse/document-render-service/blob/develop/mime_type/mimeType.go) the Content-Type needs to be text/html or the document won't be rendered but developers writing code to support a new document type wouldn't be aware of that.

> func GetMimeType(contentType string, documentType string) (mimeType *MimeType, err error) {
>	if contentType == HTML.ContentType && documentType == PDF.ContentType {
>		return PDF, nil
>	} else if contentType == HTML.ContentType && documentType == HTML.ContentType {
>		return HTML, nil
>	}
>
>	return nil, &UnsupportedMimeType{contentType, documentType}

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -48,13 +60,19 @@ public DocumentInfoResponse getAbridgedAccountsData(Transaction transaction, Str
...
         } catch (ServiceException e) {
             Map<String, Object> logMap = new HashMap<>();
             logMap.put("resource", abridgedAccountLink);
             logMap.put("accountType", accountType);
             LOG.error("Error in service layer", logMap);
             throw new HandlerException(e.getMessage(), e.getCause());
+        } catch (ParseException e) {
+            Map<String, Object> logMap = new HashMap<>();
+            logMap.put("resource", abridgedAccountLink);
+            logMap.put("accountType", accountType);
```
jsimons1989: Can you use constants instead of duplicating the logging keys please?

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
```diff
@@ -86,38 +105,75 @@ private String getAccountLink(Accounts accounts, AccountType accountsType) {
...
+    }
+
+    /**
+     * Get the period end Date from the account data
+     *
+     * @param accountData the account data
+     * @return periodEndDate The formatted periodEndDate
+     * @throws ParseException
+     */
+    private <T> String getCurrentPeriodEndOn(T accountData) throws ParseException {
```
jsimons1989: Can this return a date object? The period end on is being formatted twice event though it's only called from one place.

> return ISO_DATE_AND_TIME_FORMAT.format(ISO_DATE_FORMAT.parse(date));

> descriptionValues.put("period_end_on", RESPONSE_DISPLAY_DATE_FORMAT.format(ISO_DATE_FORMAT.
>                parse(getCurrentPeriodEndOn(accountData))));

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImplTest.java
```diff
@@ -91,16 +92,16 @@ void testErrorThrownWhenFailedAccountsHandler() throws HandlerException, Service
     @DisplayName("Tests the unsuccessful retrieval of document data due to error in Accounts handler")
     void testUnsuccessfulGetDocumentInfoExceptionFromAccountsHandler()
             throws HandlerException {
-        when(accountsHandlerMock.getAbridgedAccountsData(any(Transaction.class),  anyString())).thenThrow(new HandlerException("error"));
+        when(accountsHandlerMock.getAbridgedAccountsData(any(Transaction.class),  anyString(), anyString())).thenThrow(new HandlerException("error"));
 
-        assertThrows(HandlerException.class, () -> accountsHandlerMock.getAbridgedAccountsData(transaction, ""));
+        assertThrows(HandlerException.class, () -> accountsHandlerMock.getAbridgedAccountsData(transaction, "", MIME_TYPE));
```
jsimons1989: This is testing the mocking framework not the service

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImplTest.java
```diff
@@ -54,15 +59,35 @@ void testGetAbridgedAccountsDataFailureFromServiceLayer() throws ServiceExceptio
...
     }
 
     @Test
     @DisplayName("Tests the successful return of Abridged accounts data")
     void testGetAbridgedAccountsData() throws ServiceException, HandlerException {
         when(accountsService.getAccounts(anyString())).thenReturn(createAccountsObject());
-        when(accountsService.getAbridgedAccounts(anyString())).thenReturn(new AbridgedAccountsApi());
-        assertNotNull(accountsHandlerImpl.getAbridgedAccountsData(transaction, ACCOUNTS_RESOURCE_LINK));
+        when(accountsService.getAbridgedAccounts(anyString())).thenReturn(createCurrentPeriodAbridgedAccountObject());
+        assertNotNull(accountsHandlerImpl.getAbridgedAccountsData(transaction, ACCOUNTS_RESOURCE_LINK, MIME_TYPE));
```
jsimons1989: Can you verify the data returned please? This test will pass even if the response object contains no data.

document-generator-accounts/src/test/java/uk/gov/companieshouse/document/generator/accounts/AccountsDocumentInfoServiceImplTest.java
Mathsh: The Mock is used to check that the HandlerException is caught when the accountsHandler is called within AccountsDocumentInfoServiceImpl. As opposed to just checking that the a DocumentInfoException is caught. The reason is to ensure that if the exception is altered it is caught within the tests.

document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AccountsHandlerImpl.java
Mathsh: I've added the exception into the log

PR #46, opened 2018-10-11
Add local transaction models and fix missing accounts and abridged accounts data issue

caismith-ch:
- Add and use local transaction models over sdk models
- follow api-sdk-java example of accessing accounts and abridged accounts data
- Update api-sdk-java version

**Note**
These are temporary measures before the implementation of the private sdk.

**Resolves**: SFA-720

### DIFF
--- a/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/transaction/Resources.java
+++ b/document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/transaction/Resources.java
@@ -1,0 +1,62 @@
+package uk.gov.companieshouse.document.generator.accounts.data.transaction;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.springframework.data.mongodb.core.mapping.Field;
+
+import java.util.Date;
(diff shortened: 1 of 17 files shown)


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/transaction/Resources.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.springframework.data.mongodb.core.mapping.Field;
+
+import java.util.Date;
+import java.util.Map;
+
+public class Resources {
+
+    @Field("updated_by")
```
drosser: By using `@Field`, we're tying ourselves in to using a `MongoDB` dependency through `Spring Boot`... Do we NEED this? Can it be replaced with something else?

caismith-ch: Removed and tested, doesn't look like its necessary for what we are doing.

PR #52, opened 2018-11-08
Feature/eux 423 create mappings

ghowcroft:
added new corporate body type to api enumerations.

Jira - https://companieshouse.atlassian.net/browse/EUX-423


macreman:
looks fine for testing in an e2e but as with the api-enumerations PR - do not merge until we are 100% sure we need it.

Also, before we merge the api-enumerations commit sha will need to update to the new commit sha we get when the PR gets merged into develop

PR #71, opened 2018-11-30
Update document generator api to remove use of resourceId

Mathsh:
Update to document generator api to remove use of resourceId and change to use solely resourceUri

Updated document-generator-api to deprecated resourceId in the DocumentRequest model
Removed all use of resourceId from document-generator-api

### DIFF
--- a/document-generator-api/src/main/java/uk/gov/companieshouse/document/generator/api/models/DocumentRequest.java
+++ b/document-generator-api/src/main/java/uk/gov/companieshouse/document/generator/api/models/DocumentRequest.java
@@ -14,6 +14,7 @@ public class DocumentRequest {
     /**
      * resource id will be removed
      */
+    @Deprecated
     @JsonProperty("resource_id")
     private String resourceId;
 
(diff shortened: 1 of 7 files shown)


document-generator-api/src/main/java/uk/gov/companieshouse/document/generator/api/models/DocumentRequest.java
```diff
@@ -11,7 +11,10 @@
     @JsonProperty("resource_uri")
     private String resourceUri;
 
-    @NotNull
```
jsimons1989: Shouldn't the NOT_NULL annotation still exist until this field is removed? Even though it's deprecated it should still have the same behaviour to make sure existing code doesn't break.

Mathsh: it's not being used any more, so all reference to it has been removed from the api side. so no getters call the parameter.

PR #73, opened 2018-12-06
document generator accounts abridged handler transaction update

Mathsh:
Updated the abridgedAccountsDataHandler to obtain the transaction. 

- Updated pom to use the latest api-sdk-java version which include the changes for Links in ApiResource.
- Updated AbridgedAccountsDataHandler to use the transactionService to obtain the transaction for the required Accounts.
- Removed the previous streams that were used when obtaining the transactionLink and accountType as the Accounts object now has a Link object as opposed to a Map<String, String>.
- Corrected the date formatting as it was returning a date as a string, so no need to convert to LocalDate and then back to String.
- Updated all Junit tests
- Corrected resourceUri to resourceId when creating response.

Resolves SFA - 961


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/handler/accounts/AbridgedAccountsDataHandler.java
```diff
@@ -79,56 +76,64 @@ public DocumentInfoResponse getAbridgedAccountsData(String resourceLink, String
...
-                .filter(e -> !e.equalsIgnoreCase(LinkType.SELF.getLink()))
-                .map(AccountType::getAccountType)
-                .findFirst()
-                .orElseThrow(() -> new HandlerException("Unable to find account type in account data" +
-                        accounts.getId()));
+
+        if (accounts.getLinks().getAbridgedAccounts() != null) {
+            return AccountType.getAccountType("abridged_accounts");
+        } else {
+            throw new HandlerException("Unable to find account type in account data" +
```
BSaunders95: There needs to be a space after the word 'data' here, like `... account data " + ...`. Plus, should we use something more distinct than the Accounts `kind` for the logs? Do we have access to a unique identifier?

PR #88, opened 2019-01-25
Creditors within one year note ixbrl

JMetherell:
Add the model and mapping for creditors within one year note on the small full accounts

Resolves:
SFA-835


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/data/accounts/AccountsManager.java
```diff
@@ -148,6 +148,14 @@ public SmallFullAccountIxbrl getSmallFullAccounts(String link, Transaction trans
             handleException(e, "debtors", link);
         }
 
+        try {
+            smallFullApiData.setCreditorsWithinOneYear(apiClient.smallFull().creditorsWithinOneYear()
```
BSaunders95: Daf's just merged a PR into develop which refactors the way in which resources are retrieved - we now iterate over the small full resource links and perform a `GET` against each endpoint. You'll need to merge `develop` into this branch and use the same mechanism to set the creditors within one year resource on `smallFullApiData`.

PR #108, opened 2019-04-05
Fix accounting period dates

marcransome:
This change is required to avoid reading an incorrect `period_end_on` date when the `company_profile` document is updated after a filing is accepted. These dates are now stored in the `company_accounts` resource.

**Resolves:** SFA-1360


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/mapping/smallfull/mappers/ApiToAccountingPeriodMapper.java
```diff
@@ -0,0 +1,24 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.Mappings;
+import org.springframework.web.context.annotation.RequestScope;
+import uk.gov.companieshouse.api.model.accounts.CompanyAccountsApi;
+import uk.gov.companieshouse.document.generator.accounts.mapping.smallfull.model.ixbrl.period.Period;
+
+@RequestScope
+@Mapper(componentModel = "spring")
+@DecoratedWith(ApiToAccountingPeriodMapperDecorator.class)
+public interface ApiToAccountingPeriodMapper {
```
Mathsh: would have kept this as ApiToPeriodMapper as it was stating that it is mapping from the api to the Period model.

PR #133, opened 2019-06-18
Map CIC report data in JSON blob used to render an ixbrl document

BSaunders95:
Map CIC report data in JSON blob used to render an ixbrl document;

- Introduce `IxbrlDataWrapper` class to wrap accounts and CIC report data. This is the class which is serialized to JSON and submitted to the document render service.
- Introduce approval and statements mappers to transform CIC report data.

Required for SFA-1713


document-generator-accounts/src/main/java/uk/gov/companieshouse/document/generator/accounts/mapping/cic/mappers/ApiToStatementsMapper.java
```diff
@@ -0,0 +1,28 @@
...
+import org.mapstruct.factory.Mappers;
+import org.springframework.web.context.annotation.RequestScope;
+import uk.gov.companieshouse.api.model.accounts.cic.statements.CicStatementsApi;
+import uk.gov.companieshouse.document.generator.accounts.mapping.cic.model.Statements;
+
+@RequestScope
+@Mapper(componentModel = "spring")
+public interface ApiToStatementsMapper {
+
+    ApiToStatementsMapper INSTANCE = Mappers.getMapper(ApiToStatementsMapper.class);
```
sroberts2: Is INSTANCE needed?

PR #140, opened 2019-06-28
Feature/add charity number

ethatcher83:
Map charity number to model and add unit test
Add charity number to company report template

Resolves: PCI-102


document-generator-company-report/src/test/java/uk/gov/companieshouse/document/generator/company/report/mapping/mappers/ApiToRegistrationInformationMapperTest.java
```diff
@@ -33,6 +33,7 @@
     private String COMPANY_SUB_TYPE = "subtype";
     private LocalDate DATE_OF_CREATION = LocalDate.of(2019, 06, 06);
     private String DATE_OF_CREATION_FORMATTED = "6 June 2019";
+    private String CHARITY_NUMBER = "12345";
```
caismith-ch: This along with the other variables which are in all uppercase should be constants, for example:

`private static final String CHARITY_NUMBER = "12345";`

PR #156, opened 2019-07-29
Add recent filing history to company report

ethatcher83:
-Create recent filing history model
-Create recent filing history mapper
-Add recent filing history section to company report template
-Create recent filing history service
-Update Company report data handler
-Create unit tests for recent filing history

Resolves: PCI-53 
Resolves: PCI-63


document-generator-company-report/src/main/java/uk/gov/companieshouse/document/generator/company/report/mapping/mappers/recentfilinghistory/ApiToRecentFilingHistoryMapper.java
```diff
@@ -67,6 +77,17 @@ private DateTimeFormatter getFormatter() {
...
     private String populateParameters(Object description, Map<String, Object> parameters) {
         StrSubstitutor sub = new StrSubstitutor(parameters, "{", "}");
+
+        for (String parameterKey : parameters.keySet()) {
+
+            if (parameterKey.contains("_date")) {
+
+                LocalDate localDate = LocalDate.parse(parameters.get(parameterKey).toString());
+
+                parameters.replace(parameterKey, localDate.format(getFormatter()));
```
Mathsh: Should probably use replace(K key ,V oldValue, V newValue) in order to check that the key and value are the same as the old one. before updating to the new value. As we could potentially have two of the same keys and we update the wrong one.

PR #191, opened 2019-10-16
Add logic to check if active exemption already true to not overwrite …

ethatcher83:
Check if active exemption is already true, if it is, not to overwrite as false when checking all types of exemptions.

Resolves: PCI-235


document-generator-company-report/src/main/java/uk/gov/companieshouse/document/generator/company/report/mapping/mappers/exemptions/ApiToExemptionsMapper.java
```diff
@@ -27,33 +27,41 @@ void checkActiveExemption(ExemptionsApi exemptionsApi, @MappingTarget Exemptions
         if(exemptionsApi.getDisclosureTransparencyRulesChapterFiveApplies() != null) {
              disclosureTransparencyRulesChapterFiveAppliesList = exemptionsApi.getDisclosureTransparencyRulesChapterFiveApplies().getItems();
 
-            exemptions.setActiveExemption(false);
+             if(!exemptions.isActiveExemption()) {
+                 exemptions.setActiveExemption(false);
```
caismith-ch: Why are we setting `isActiveExemption` to `false` after checking if it is `false`?

PR #239, opened 2020-11-03
Resolve issue with no director name

faisal-amin82:
- Resolves issue with Loans to Directors resource where director name is now optional and not provided due to gdpr changes

Resolves: BI-5799


ch-tstrickland (approved):
Looks good to me.
The commit message would have been more useful if it included the text in the PR message:
Original commit:
Resolve issue with no director name

What issue? You don't say.
The commit could have read:
Make Director Name optional
In line with the GDPR, Director Name should be optional.
These changes make the mapper cope with a possible blank director name.

This would be much more useful to someone having to work out why these changes were made. They will not read the PR text. They will not have access to the JIRA. All they have is the Git log. Be nice to future developers that have to cope with our code: write useful commit messages.

PR #266, opened 2022-01-21
security upgrades jan 22

SteveHicksCH:
Update the following libraries for security warnings: snakeyaml, httpclient, commons-io, gson

Also moved any common property (that was updated) to the parent pom

Ref: BI-9903

jar checked OK for new versions

$ jar -tf document-generator-unversioned.jar | grep snakeyaml
BOOT-INF/lib/snakeyaml-1.26.jar
$  jar -tf document-generator-unversioned.jar | grep  httpclient
BOOT-INF/lib/httpclient-4.5.13.jar
$  jar -tf document-generator-unversioned.jar | grep  commons-io
BOOT-INF/lib/commons-io-2.7.jar
$  jar -tf document-generator-unversioned.jar | grep  gson
BOOT-INF/lib/gson-2.8.9.jar

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ bin/
 .idea/
 *.iml
 
+# VSCode
+.vscode/
+
 # Mac files
 .DS_Store
 ._*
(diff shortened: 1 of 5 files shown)


ch-tstrickland (approved):
Approved - but changes to WS & gitignore could have been in a separate PR.

PR #326, opened 2024-10-15
CC-1546 - Remove broken linked CDN assets

ch-ncolley:
Work has [previously been done to make these assets redundant](
https://github.com/companieshouse/document-generator/pull/190) and given they've not been loading at all we can remove them rather than ensuring they're up-to-date.


ch-code-analysis:
CI: No security warnings found

PR #352, opened 2025-08-27
Use chs-internal-apikey instead of chs-api-key to call services.

myadavch:
Make changes to use `chs-internal-api-key` instead of `chs-api-key`.


ch-code-analysis:
CI: Security warnings found!

modhasanjay (approved):

